PR #9, opened 2020-10-06
Fix Tag

zawachte-msft:
moc-sdk-for-go pointing that this tag is breaking repos that need moc at 10.4.


JocelynBerrendonner (approved):
LGTM

nwoodmsft:
Thanks @zawachte-msft

PR #13, opened 2020-10-20
Adding compute.VirtualMachine GetByComputerName

nwoodmsft:
Wrapper function to make it easier to "Get" a virtual machine by it's computer name. More complex queries are already supported by Query(). 

GetByComputerName will be invoked by CSI and MOC Provider etc.

Example usage:

`vms, err := vmclient.GetByComputerName(ctx, group, "nickcluster-md-0-nzdkw")`

### DIFF
--- a/services/compute/virtualmachine/client.go
+++ b/services/compute/virtualmachine/client.go
@@ -5,6 +5,7 @@ package virtualmachine
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"time"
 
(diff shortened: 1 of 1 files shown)


madhanrm (approved):
Are you expecting multiple Vms with same computer names?

nwoodmsft:
I was thinking we might support wildcards here, so there would be potential for multiple matching VMs to be returned.

PR #14, opened 2020-10-21
Query should use JSON marshaler

nwoodmsft:
Using the YAML marshaler is not safe in this case as some property names within our structs have a mix of uppercase/lowercase. Converting to YAML bytes and then back will cause data loss (the mixed case properties will be dropped). JSON handles this case much better.

### DIFF
--- a/services/compute/virtualmachine/wssd.go
+++ b/services/compute/virtualmachine/wssd.go
@@ -101,12 +101,13 @@ func (c *client) Query(ctx context.Context, group, query string) (*[]compute.Vir
 		return nil, err
 	}
 
-	filteredBytes, err := config.MarshalOutput(*vms, query, "yaml")
+	filteredBytes, err := config.MarshalOutput(*vms, query, "json")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


services/compute/virtualmachine/wssd.go
```diff
@@ -101,12 +101,13 @@ func (c *client) Query(ctx context.Context, group, query string) (*[]compute.Vir
 		return nil, err
 	}
 
-	filteredBytes, err := config.MarshalOutput(*vms, query, "yaml")
+	filteredBytes, err := config.MarshalOutput(*vms, query, "json")
 	if err != nil {
 		return nil, err
 	}
+	fmt.Printf("ITS: %s\n", string(filteredBytes))
```
zawachte-msft, line 108: ITS ALIVE https://www.youtube.com/watch?v=xos2MnVxe-c&ab_channel=crawlfan

zawachte-msft (approved):

services/compute/virtualmachine/wssd.go
nwoodmsft: Thanks! lol...golang debugging FTW

PR #16, opened 2020-11-03
Changing Certificate type to string

apluchik:
Changing the certificate types to string from []byte.
related to https://github.com/microsoft/moc/pull/14

### DIFF
--- a/services/security/security.go
+++ b/services/security/security.go
@@ -243,7 +243,7 @@ type Certificate struct {
 	// X509Thumbprint - READ-ONLY; Thumbprint of the certificate. (a URL-encoded base64 string)
 	X509Thumbprint *string `json:"x5t,omitempty"`
 	// Cer - CER contents of x509 certificate.
-	Cer *[]byte `json:"cer,omitempty"`
+	Cer *string `json:"cer,omitempty"`
 	// Type - The content type of the certificate
 	Type *string `json:"contentType,omitempty"`
 	// Attributes - The certificate attributes.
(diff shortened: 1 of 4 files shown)


services/security/security.go
```diff
@@ -243,7 +243,7 @@ type Certificate struct {
 	// X509Thumbprint - READ-ONLY; Thumbprint of the certificate. (a URL-encoded base64 string)
 	X509Thumbprint *string `json:"x5t,omitempty"`
 	// Cer - CER contents of x509 certificate.
```
madhanrm, line 245: can you also update the comment saying this is base64 encoded string

services/security/security.go
```diff
@@ -268,8 +268,8 @@ type Identity struct {
 	Type *string `json:"type,omitempty"`
 	// Token
 	Token *string `json:"token,omitempty"`
-	// Certificate
-	Certificate *[]byte `json:"certificate,omitempty"`
+	// Certificates
+	Certificate *string `json:"certificates,omitempty"`
```
madhanrm, line 272: is this the list of certificates that are held in map?

apluchik: This is the certificate passed in to the cloud agent, the map of client to cert is stored in the security.proto Identity struct.

services/security/security.go
apluchik: Done.

JocelynBerrendonner (approved):
LGTM

PR #28, opened 2021-02-16
Cloud SDK changes for bare metal machine

davutkarabay:
Cloud SDK changes for bare metal machine entity.

Note: go.mod and go.sum files will be updated after the protocol buffer PR is merged. Currently the changes are built against private MOC.

### DIFF
--- a/pkg/client/compute.go
+++ b/pkg/client/compute.go
@@ -9,7 +9,7 @@ import (
 	compute_pb "github.com/microsoft/moc/rpc/cloudagent/compute"
 )
 
-// GetGalleryImageClient returns the virtual machine client to comminicate with the wssd agent
+// GetGalleryImageClient returns the virtual machine client to communicate with the wssd agent
 func GetGalleryImageClient(serverAddress *string, authorizer auth.Authorizer) (compute_pb.GalleryImageAgentClient, error) {
 	conn, err := getClientConnection(serverAddress, authorizer)
 	if err != nil {
@@ -19,7 +19,7 @@ func GetGalleryImageClient(serverAddress *string, authorizer auth.Authorizer) (c
 	return compute_pb.NewGalleryImageAgentClient(conn), nil
 }
 
-// GetVirtualMachineClient returns the virtual machine client to comminicate with the wssd agent
+// GetVirtualMachineClient returns the virtual machine client to communicate with the wssd agent
 func GetVirtualMachineClient(serverAddress *string, authorizer auth.Authorizer) (compute_pb.VirtualMachineAgentClient, error) {
 	conn, err := getClientConnection(serverAddress, authorizer)
 	if err != nil {
@@ -29,7 +29,7 @@ func GetVirtualMachineClient(serverAddress *string, authorizer auth.Authorizer) 
 	return compute_pb.NewVirtualMachineAgentClient(conn), nil
 }
 
-// GetVirtualMachineScaleSetClient returns the virtual machine client to comminicate with the wssd agent
+// GetVirtualMachineScaleSetClient returns the virtual machine client to communicate with the wssd agent
 func GetVirtualMachineScaleSetClient(serverAddress *string, authorizer auth.Authorizer) (compute_pb.VirtualMachineScaleSetAgentClient, error) {
 	conn, err := getClientConnection(serverAddress, authorizer)
 	if err != nil {
@@ -39,7 +39,7 @@ func GetVirtualMachineScaleSetClient(serverAddress *string, authorizer auth.Auth
 	return compute_pb.NewVirtualMachineScaleSetAgentClient(conn), nil
 }
 
-// GetBareMetalMachineClient returns the bare metal machine client to comminicate with the wssd agent
+// GetBareMetalMachineClient returns the bare metal machine client to communicate with the wssd agent
 func GetBareMetalMachineClient(serverAddress *string, authorizer auth.Authorizer) (compute_pb.BareMetalMachineAgentClient, error) {
 	conn, err := getClientConnection(serverAddress, authorizer)
 	if err != nil {
--- a/pkg/client/compute.go
+++ b/pkg/client/compute.go
@@ -38,3 +38,13 @@ func GetVirtualMachineScaleSetClient(serverAddress *string, authorizer auth.Auth
 
 	return compute_pb.NewVirtualMachineScaleSetAgentClient(conn), nil
 }
+
+// GetBareMetalMachineClient returns the bare metal machine client to comminicate with the wssd agent
+func GetBareMetalMachineClient(serverAddress *string, authorizer auth.Authorizer) (compute_pb.BareMetalMachineAgentClient, error) {
+	conn, err := getClientConnection(serverAddress, authorizer)
+	if err != nil {
+		log.Fatalf("Unable to get BareMetalMachineClient. Failed to dial: %v", err)
+	}
+
+	return compute_pb.NewBareMetalMachineAgentClient(conn), nil
+}
--- a/services/compute/baremetalmachine/baremetalmachine.go
+++ b/services/compute/baremetalmachine/baremetalmachine.go
@@ -1,0 +1,379 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache v2.0 License.
+
+package baremetalmachine
+
+import (
+	"github.com/microsoft/moc/pkg/convert"
+	"github.com/microsoft/moc/pkg/errors"
+
+	"github.com/microsoft/moc-sdk-for-go/services/compute"
+
+	"github.com/microsoft/moc/pkg/status"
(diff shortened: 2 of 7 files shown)


pkg/client/compute.go
```diff
@@ -38,3 +38,13 @@ func GetVirtualMachineScaleSetClient(serverAddress *string, authorizer auth.Auth
 
 	return compute_pb.NewVirtualMachineScaleSetAgentClient(conn), nil
 }
+
+// GetBareMetalMachineClient returns the bare metal machine client to comminicate with the wssd agent
```
jiria, line 42: nit: typo in communicate

services/compute/baremetalmachine/baremetalmachine.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+	hardwareConfig, err := c.getWssdBareMetalMachineHardwareConfiguration(bmm)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to get Hardware Configuration")
+	}
+	securityConfig, err := c.getWssdBareMetalMachineSecurityConfiguration(bmm)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to get Security Configuration")
+	}
+	osconfig, err := c.getWssdBareMetalMachineOSConfiguration(bmm.OsProfile)
```
jiria, line 35: nit: osconfig casing

services/compute/baremetalmachine/baremetalmachine.go
```diff
@@ -0,0 +1,379 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to get OS Configuration")
+	}
+
+	networkConfig, err := c.getWssdBareMetalMachineNetworkConfiguration(bmm.NetworkProfile)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to get Network Configuration")
+	}
+
+	bmmOut := wssdcloudcompute.BareMetalMachine{
```
jiria, line 45: Are you not outputting nodeName on purpose?

services/compute/baremetalmachine/baremetalmachine.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+	return *s.Name, nil
+}
+
+func (c *client) getWssdBareMetalMachineStorageConfigurationDisk(s *compute.BareMetalMachineDisk) (*wssdcloudcompute.BareMetalMachineDisk, error) {
+	if s.Name == nil {
+		return nil, errors.Wrapf(errors.InvalidInput, "Name is missing in BareMetalMachineDisk")
+	}
+	if s.DiskSizeGB == nil {
+		return nil, errors.Wrapf(errors.InvalidInput, "Disk Size is missing in BareMetalMachineDisk ")
```
jiria, line 114: nit: extra space in the error message at the end

services/compute/baremetalmachine/baremetalmachine.go
```diff
@@ -0,0 +1,379 @@
...
+			Statuses:          c.getBareMetalMachineStatuses(bmm),
+			StorageProfile:    c.getBareMetalMachineStorageProfile(bmm.Storage),
+			HardwareProfile:   c.getBareMetalMachineHardwareProfile(bmm),
+			SecurityProfile:   c.getBareMetalMachineSecurityProfile(bmm),
+			OsProfile:         c.getBareMetalMachineOSProfile(bmm.Os),
+			NetworkProfile:    c.getBareMetalMachineNetworkProfile(bmm.Network),
+			Host:              c.getBareMetalMachineHostDescription(bmm),
+		},
+		Version:  &bmm.Status.Version.Number,
+		Location: &bmm.LocationName,
```
jiria, line 260: This seems to be missing GroupName and NodeName. Is that on purpose?

services/compute/baremetalmachine/baremetalmachine.go
davutkarabay: This is mostly cloned from VM where it was missing. Still learning, but this is _In_ path to the cloud agent where it may not be possible to know the node agent's name. I think it is something to be resolved in the cloud agent (currently work in progress) and set in the _Out_ path which we do (see your other comment).

services/compute/baremetalmachine/baremetalmachine.go
davutkarabay: NodeName is contained in Host. This is how it is for the VM and I followed. GroupName is not defined in the SDK. Note that group is a param to the API. I believe for a VM, both location and group have to be present. I assummed the same for a bare metal machine.

services/compute/baremetalmachine/baremetalmachine.go
jiria: You are right, this should not be settable by the SDK.

services/compute/baremetalmachine/baremetalmachine.go
jiria: Sounds good, lets keep it symmetric for now.

jiria (approved):

jiria:
Btw, CI is failing with:

pkg/client/compute.go:43:84: undefined: compute.BareMetalMachineAgentClient
pkg/client/compute.go:49:9: undefined: compute.NewBareMetalMachineAgentClient

jiria (changes requested):
Please fix the CI failures.

services/compute/baremetalmachine/baremetalmachine.go
davutkarabay: I may have caused confusion. In my proposed design, bare metal machines are created (the first time on-boarding) after the node creation. So at that time the setup script knows the node name and should pass it to the BMM. Otherwise the cloud agent does not know the node. For VMs, passing the node name is optional because the cloud agent picks a node based on VM size requirements, which I think makes sense. For bare metal, the node agent host is the BMM host. So, you were right, we should set the nodeName here.

jiria: Interesting, I was actually thinking about it the other way around. We first have BMM in cloud agent. This is how cloud agent can talk to BMC for example (as defined in the BMM, eventually, provider specific). Then CA through the BMM provider provisions the BMM and deploys Node Agent. Only then there would be Node entity, which can be added into the BMM entity. Curious about your thoughts on this.

davutkarabay: How does BMM entity in CA talk to the BMM entity (or any entity) in node agent? I believe it is done via the node entity in CA. See GetNodeVirtualMachineClient() for VM reference. I was thinking I'd create a corresponding GetNodeBareMetalMachineClient(). In my mind the steps were following:

Via setup script at setup time:
1. Configure cloud agent
2. Configure node agent
3. Create location
4. Create group
5. Create node
6. Create BMM. This is on-boarding the BM node to the cloud. Only BM node capabilities are passed e.g., OS/storage/network info.

Via BM controller at a later time:
7. Update BMM. This is to add the BM node to a target cluster. Cloud-init (and possibly a target image ref) is passed.

jiria: Step 2 expects that you have OS already deployed on that host, right? (Not for step 1, since Cloud Agent can be running elsewhere.) Here is my counter point scenario:
1) Say I have done what you proposed on a single host
2) I want to add another BM host; there is no OS, so first I need to onboard the host to MOC to let MOC drive the OS deployment
3) To do that, I create the BMM entity and pass BMC creds to it
4) MOC BMM entity creation is passed to the BMM provider, which then goes and deploys OS onto the host through BMC (no Node Agent yet), and initializes the host with cloud-init (this can deploy and configure Node Agent)
5) With Node Agent ready, additional operations can be performed through NodeAgent.

Now this is not what we are doing for March; for March, we will deploy OS upfront through Robin and so your flow will work. But I just wanted to point out that we might need to refactor that post March.

Here is my current thinking on the March flow: http://aka.ms/robin-march
Here is my current thinking on post March: https://aka.ms/robin-provider-no-dashboard

If you look at the second flow and the Add host to ECF section, it illustrates what I was trying to put in that counter point scenario. If you see anything wrong with the flows or have any other thoughts or feedback, would be happy to discuss and augment.

davutkarabay: I see - thanks. Yes, my steps assumed the OS is pre-deployed. The flow makes sense mostly. I have some questions, but will ask them privately.

davutkarabay:
> Btw, CI is failing with:
> 
> pkg/client/compute.go:43:84: undefined: compute.BareMetalMachineAgentClient
> pkg/client/compute.go:49:9: undefined: compute.NewBareMetalMachineAgentClient

These were are defined in the other PR for the MOC, which is merged now. I updated the go.mod to reference the MOC changes which resolved the issue.

jiria (approved):

madhanrm (approved):

PR #29, opened 2021-02-18
Adding Identity fields for node to cloud agent login

apluchik:
Adding new fields to Identity and IdentityProperties structs for LoginConfig creation.


services/security/identity/identity.go
```diff
@@ -9,15 +9,29 @@ import (
...
 	"github.com/microsoft/moc/pkg/status"
 	wssdcloudsecurity "github.com/microsoft/moc/rpc/cloudagent/security"
+	wssdcloudcommon "github.com/microsoft/moc/rpc/common"
 )
 
 func getIdentity(id *wssdcloudsecurity.Identity) *security.Identity {
+	clitype := security.ExternalClient
+	if id.ClientType == wssdcloudcommon.ClientType_CONTROLPLANE {
+		clitype = security.ControlPlane
+	} else if id.ClientType == wssdcloudcommon.ClientType_EXTERNALCLIENT {
```
madhanrm, line 19: ExternalClient and Node is not same right?

apluchik: good catch. This should be checking for ClientType_NODE, and external client should be default (external client will have faster expiry/more restrictions in the future).

madhanrm (approved):

services/security/security.go
```diff
@@ -283,6 +291,14 @@ type CertificateRequest struct {
 type IdentityProperties struct {
 	// State - State
 	Statuses map[string]*string `json:"statuses"`
+	// CloudAgent FQDN
+	CloudFqdn *string `json:"cloudfqdn,omitempty"`
+	// CloudAgent port
+	CloudPort *int32 `json:"cloudport,omitempty"`
+	// CloudAgent authentication port
+	CloudAuthPort *int32 `json:"cloudauthport,omitempty"`
```
cleonard-git, line 299: Ports can be uint . Any specific reason for using signed int?

apluchik: Good point. It's added as int32 to stay consistent with existing definitions for Ports our repo, as well as the azure-sdk-for-go repo.

PR #34, opened 2021-03-16
Renew routine for ca-signed certs

raghavendra-nataraj:


services/security/authentication/casigned/casigned.go
```diff
@@ -37,6 +44,47 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
...
+func RenewRoutine(group, server string) {
+	renewalAttempt := 0
+	// Waiting for a few seconds to avoid spamming short-lived sdk user
+	time.Sleep(time.Second * 5)
+	for {
+		wssdConfig := auth.WssdConfig{}
+		err := marshal.FromJSONFile(auth.GetWssdConfigLocation(), &wssdConfig)
+		if err != nil {
+			fmt.Printf("Failed to open config file in location %s: %v\n", auth.GetWssdConfigLocation(), err)
+			panic("Failed to open config file")
```
JocelynBerrendonner, line 56: Can we either return an error, or log the error and return gracefully instead of panicking? Keep in mind that panic will crash the app that uses the SDK. This typically is used by attackers to generate denial of service attacks. So unless the state of the program is unrecoverable (i.e. bad memory corruption), we should try to be graceful.

Here we seem to be in a callback, so I understand we don't have the ability to return the error to the caller, but we should log the error and just return to the caller. If we fail to renew CA certs, the user will notice it very quickly.

services/security/authentication/casigned/casigned.go
```diff
@@ -37,6 +44,47 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
...
+		}
+
+		sleepTime, err := renewTime(wssdConfig.ClientCertificate)
+		if err != nil {
+			fmt.Printf("Failed to find sleep time for cert %v \n", err)
+			panic("Failed to find renew time for certificate")
```
JocelynBerrendonner, line 62: Same here (also applies to other locations below)

services/security/authentication/casigned/casigned.go
```diff
@@ -37,6 +44,47 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
...
+		}
+		fmt.Printf("Waiting for %v to renew cert\n", sleepTime)
```
JocelynBerrendonner, line 64: [nitpick] the problem with Printf statements is that if a command line tool uses the SDK, its output will suddenly be "polluted" with all the logging coming from the SDK. Is there a way we could pass a logger from the Client to the routine? Feel free to push back and we can handle this later.

raghavendra-nataraj: I thought of introducing lambda for logging, but felt printing would be simpler solution as the renewal is not a very frequent operation. I can introduce loggers if needed in another iteration

services/security/authentication/casigned/casigned.go
```diff
@@ -37,6 +44,47 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
...
 
+func RenewRoutine(group, server string) {
+	renewalAttempt := 0
+	// Waiting for a few seconds to avoid spamming short-lived sdk user
+	time.Sleep(time.Second * 5)
+	for {
+		wssdConfig := auth.WssdConfig{}
+		err := marshal.FromJSONFile(auth.GetWssdConfigLocation(), &wssdConfig)
+		if err != nil {
+			fmt.Printf("Failed to open config file in location %s: %v\n", auth.GetWssdConfigLocation(), err)
```
pradipd, line 55: +1 To what Jocelyn says.
Don't panic and don't use fmt.Printf.  
Return the error you are printing out and have the caller handle it.

services/security/authentication/casigned/casigned.go
```diff
@@ -37,6 +44,47 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
 	return &response.Token, nil
 }
 
+func RenewRoutine(group, server string) {
+	renewalAttempt := 0
+	// Waiting for a few seconds to avoid spamming short-lived sdk user
+	time.Sleep(time.Second * 5)
```
pradipd, line 50: What is a "short-lived" sdk user?  
What is the sleep() trying to do?

raghavendra-nataraj: This is to avoid spamming the sdk client like mocctl

services/security/authentication/casigned/casigned.go
raghavendra-nataraj: The caller is unaware of the renewal process, it is in a background thread that the renewal is happening

services/security/authentication/casigned/casigned.go
pradipd: I think you need the LoginWithConfig to take in a ctx so you can log these events.  Our life will be miserable if we do not log these.
LoginWithConfig(ctx context.Context, ....)  and pass that to go RenewRouting(ctx context.Context, ...)
Then you will be able to log all your printfs

services/security/authentication/casigned/casigned.go
```diff
@@ -37,6 +44,47 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
...
+				panic("Failed to renew certificate")
+			}
+			if errors.IsExpired(err) {
+				fmt.Printf("Certificate Expired %v \n", err)
+				panic("Certificate has expired")
+			}
+			renewalAttempt += 1
+			// Wait for 30 seconds before next renewal attempt
+			time.Sleep(time.Second * 30)
+			continue
```
madhanrm, line 80: 30 seconds is too less.
Everything should be a factor of the cert validity time.

services/security/authentication/casigned/casigned.go
```diff
@@ -37,6 +44,47 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
...
+				fmt.Printf("Failed to renew cert to %v \n", err)
+				panic("Failed to renew certificate")
+			}
+			if errors.IsExpired(err) {
+				fmt.Printf("Certificate Expired %v \n", err)
+				panic("Certificate has expired")
+			}
+			renewalAttempt += 1
+			// Wait for 30 seconds before next renewal attempt
+			time.Sleep(time.Second * 30)
```
pradipd, line 79: Doesn't this time need to be based of the sleepTime?
And why cap it at 5?  Why not keep letting it run (with a delay in between each attempt) till it expires?

services/security/authentication/casigned/casigned.go
```diff
@@ -37,6 +44,47 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
...
+			panic("Failed to find renew time for certificate")
+		}
+		fmt.Printf("Waiting for %v to renew cert\n", sleepTime)
+		time.Sleep(sleepTime)
+		fmt.Printf("Attempting to renew certificate\n")
+		err = auth.RenewCertificates(server, auth.GetWssdConfigLocation())
+		if err != nil {
+			if renewalAttempt > 5 {
+				fmt.Printf("Failed to renew cert to %v \n", err)
+				panic("Failed to renew certificate")
```
madhanrm, line 71: Once we fix the sleep timer, we dont have to panic here.
Panic should be done only if the cert expired, and we want to force app to login again to get a new certificate.
Remember, in this context the cert is still valid and can communicate with MOC

services/security/authentication/casigned/casigned.go
pradipd: It would be awesome if you could display a bomb timer in the logs and play some terrifying music. 
Seriously, the time remaining before expiry might be nice.
Or something similar (the current time, and the time it expires)
Whatever is easier

services/security/authentication/casigned/casigned.go
```diff
@@ -61,9 +109,33 @@ func (c *client) LoginWithConfig(group string, loginconfig auth.LoginConfig) (*a
 	accessFile.ClientCertificateType = auth.CASigned
 	accessFile.IdentityName = loginconfig.Name
 	auth.PrintAccessFile(accessFile)
+	go RenewRoutine(group, c.cloudFQDN)
```
zawachte-msft, line 112: How do we prevent multiple goroutines from happening? Do we need a lock here?

raghavendra-nataraj: Good point. multiple logins could trigger multiple go routines.

services/security/authentication/casigned/casigned.go
```diff
@@ -37,8 +46,49 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
...
+			span.Log("Failed while calculating certificate renew time %v \n", err)
+			return
+		}
+		span.Log("Waiting for %v to renew cert\n", sleepTime)
+		time.Sleep(sleepTime)
+		span.Log("Attempting to renew certificate\n")
+		err = auth.RenewCertificates(server, auth.GetWssdConfigLocation())
+		if err != nil {
+			if errors.IsExpired(err) {
+				span.Log("Certificate Expired %v", err)
```
madhanrm, line 75: Do we have context here to perform an auto login?

services/security/authentication/casigned/casigned.go
madhanrm: go routine should be triggered only if we created a certificate. we could exit before creating a certificate, if it already exists

raghavendra-nataraj: We do not know the state of the old certificate that exists

services/security/authentication/casigned/casigned.go
raghavendra-nataraj: We do have but the token we have could be expired most likely as the token validity is likely less than the cert validity

madhanrm: If auto rotation was done, the possibility of token expiry is even reduced and trying to login back should get the cases of panic even lesser

madhanrm: lets handle that is next PR

madhanrm (approved):

services/security/authentication/casigned/casigned.go
pradipd: So, you are avoiding renewing the certificate for mocctl?
is the expectation that mocctl will finish the command in 5 seconds and will thus exit the program?

raghavendra-nataraj: yes, I assume the login would take tell than that.

pradipd: what if the mocctl command runs for more than 5 seconds?  E.g. when you create a vm it can take more than 5 seconds.  Or if we are running on a overloaded computer?
Waiting 5 seconds is not a gaurantee.  It would be better to have the client pass in something that says it does not need to renew.

raghavendra-nataraj: This is only for the login api. VM creation would not invoke this. Yes, adding the parameter for renew would be better. I'll take it up next PR.

pradipd: ok.

pradipd (approved):

PR #36, opened 2021-03-23
Fix the order of processing VmConfigContainerName so it is picked wit…

madhanrm:
…h datadisk is not empty


JocelynBerrendonner (approved):
LGTM

PR #38, opened 2021-03-26
Adding Role and RoleAssignment resources

apluchik:
Adding Role and RoleAssignment resources

Depends on :
https://github.com/microsoft/moc/pull/47


services/security/providerTypes.go
```diff
@@ -0,0 +1,66 @@
...
+func GetProviderType(pbProvider pbcom.ProviderType) ProviderType {
+	providerInt := int32(pbProvider)
+	value, found := pbcom.ProviderType_name[providerInt]
+	if !found {
+		return AnyProviderType // Not found, return AnyProvider
+	}
+	return ProviderType(value)
+}
+
+func GetWssdProviderType(providerType ProviderType) (pbcom.ProviderType, error) {
```
madhanrm, line 20: Try to use Moc instead of Wssd in new code.

services/security/security.go
```diff
@@ -160,12 +160,13 @@ const (
...
 
-type ClientType string
+type Operation string
 
 const (
-	ControlPlane   ClientType = "ControlPlane"
-	ExternalClient ClientType = "ExternalClient"
-	Node           ClientType = "Node"
+	ReadAccess   Operation = "read"
+	WriteAccess  Operation = "write"
```
madhanrm, line 167: How do I differentiate between Creating a Group Vs Creating Resource in a Group.
What access should be available on Parent to allow this ie 
for group create, what permission in Location is required.

services/security/providerTypes.go
apluchik: Changed wssd references to moc.

services/security/security.go
apluchik: For creating a group, a role with a permission containing write/all and a providerType of any or group would allow creation of a group. Assigning this role using a scope under a location (or not bound by any scope), will allow a client to create a group.
To create a resource in a group we can have a role with permissions to create the resource, and assign it using a scope that is specifying a specific group.

madhanrm: Definitely need a huge md for this. Please try to get a md file after enabling the feature.

madhanrm (approved):

PR #51, opened 2021-05-06
CVE fixes

pradipd:


madhanrm (approved):

madhanrm:
are you going to revendor all the clients to pick this CVE?

pradipd:
Possibly.  If this package is the only reason the client has the websocket CVE, then I'll revendor.
However, if other packages are also bringing in this CVE, then revendoring isn't going to make it go away.  
I'll have to replace.

PR #54, opened 2021-05-24
passing powerstate from protobuf to moc-sdk

raghavendra-nataraj:


services/compute/compute.go
```diff
@@ -233,6 +243,8 @@ type VirtualMachineProperties struct {
 	VmType VMType `json:"vmType,omitempty"`
 	// Disable High Availability
 	DisableHighAvailability *bool `json:"disableHighAvailability,omitempty"`
+	// PowerState
+	PowerState PowerState `json:"powerstate,omitempty"`
```
madhanrm, line 247: are we making this a property now?
How is it represented in Azure SDK

madhanrm: Do we still need this ?

PR #64, opened 2021-07-26
moc-sdk changes for sfs download

cleonard-git:


services/compute/galleryimage/client.go
```diff
@@ -39,10 +40,30 @@ func (c *GalleryImageClient) Get(ctx context.Context, location, name string) (*[
...
 }
 
 // Delete methods invokes delete of the compute resource
 func (c *GalleryImageClient) Delete(ctx context.Context, location, name string) error {
 	return c.internal.Delete(ctx, location, name)
 }
+
+// UploadImageFromLocal   methods invokes  UploadImageFromLocal  on the client
+func (c *GalleryImageClient) UploadImageFromLocal(ctx context.Context, location, imagePath, name string, compute *compute.GalleryImage) (*compute.GalleryImage, error) {
+	compute.SourceType = "local"
```
madhanrm, line 54: can you define "local", "sfs" as enums in moc and reuse them. It is not dev friendly to hard code these here.

cleonard-git: yeah I had that in the node agent code.  Brought it moc-sdk as well.

pkg/constant/constant.go
```diff
@@ -11,4 +11,8 @@ const (
 	DefaultServerContextTimeout          = 10 * time.Minute
 	CertificateValidityThreshold float64 = (30.0 / 100.0)
 	RenewalBackoff               float64 = (2.0 / 100.0)
+	SFSSource                            = "sfs"
```
madhanrm, line 14: We could define these as .proto enums and get it auto generated, and used across all the layers.

cleonard-git: Done . https://github.com/microsoft/moc/pull/82

madhanrm (approved):

PR #65, opened 2021-07-26
Fix WindowsConfiguration nil check

aeweston-msft:
## Changes introduced by this PR
* Fix nil check so that WinRM configuration will be copied


JocelynBerrendonner (approved):
LGTM

madhanrm (approved):

PR #70, opened 2021-08-11
moc-sdk : null ptr check for SourceType for backward compatibility

cleonard-git:


madhanrm (approved):

cleonard-git:
> Can we make sure we test everything end to end and then raise PRs.
> That way we dont have to keep making sdk changes.

Testing with wssdtest was done with my changes  which rested the feature , so this was not caught earlier. Vanilla wssdtest with  latest tags from all repos should be tested. Will keep this scenario in mind next time.

PR #80, opened 2021-10-08
Add taints and tolerations

cwize1:


services/cloud/cloud.go
```diff
@@ -65,6 +66,9 @@ type NodeProperties struct {
 	AuthorizerPort *int32 `json:"authorizerPort,omitempty"`
 
 	Certificate *string `json:"certificate,omitempty"`
+
+	// Taints that restricts which VMs can run on this node.
+	Taints []common.Taint `json:"taints"`
```
bfjelds, line 71: omitempty?

go.mod
```diff
@@ -19,3 +19,5 @@ replace (
 	github.com/Azure/go-autorest v11.1.2+incompatible => github.com/Azure/go-autorest/autorest v0.10.0
 	github.com/miekg/dns => github.com/miekg/dns v1.1.25
 )
+
+replace github.com/microsoft/moc => ../moc
```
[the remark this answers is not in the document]
cwize1: Just local testing. Thanks for the catch.

services/common/common.go
```diff
@@ -0,0 +1,29 @@
...
+	// How the taint's value is handled.
+	Operator TolerationOperator `json:"operator"`
+	// The key of the taint to tolerate.
+	Key string `json:"key"`
+	// The value to match against the taint's value.
+	// This field is ignored if 'operator' is set to 'Exists'.
+	Value string `json:"value"`
+	// If true, toleration must match a taint. If false, taint may be present
+	// but is not required.
+	Required bool `json:"required,omitempty"`
```
vincepnguyen, line 28: If its not set, empty field, is it required or not?

services/compute/compute.go
```diff
@@ -277,6 +278,8 @@ type VirtualMachineProperties struct {
 	VmType VMType `json:"vmType,omitempty"`
 	// Disable High Availability
 	DisableHighAvailability *bool `json:"disableHighAvailability,omitempty"`
+	// Allows VM to run on nodes that have been tainted for a specific workload.
+	NodeTolerations []common.Toleration `json:"nodetolerations"`
```
bfjelds, line 282: omitempty?

services/common/common.go
cwize1: If it isn't set, this it will default to `false` which I am happy with.

PR #83, opened 2021-11-11
Return vm sizes

acraghav:
Expose an API to return VM sizes from moc-pkg.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Azure/go-autorest/autorest v0.9.0
 	github.com/Azure/go-autorest/autorest/date v0.2.0
 	github.com/google/uuid v1.2.0
-	github.com/microsoft/moc v0.10.15-alpha.1
+	github.com/microsoft/moc v0.10.16-alpha.2
 	github.com/spf13/viper v1.7.1
 	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013 // indirect
 	google.golang.org/grpc v1.27.1
--- a/services/compute/vmSizes.go
+++ b/services/compute/vmSizes.go
@@ -4,6 +4,7 @@
 package compute
 
 import (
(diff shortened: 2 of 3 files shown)


services/compute/vmSizes.go
```diff
@@ -445,3 +453,26 @@ func GetVirtualMachineSizes() (vmsizes *[]VirtualMachineSizeTypes) {
 	vmsizes = &tmp
 	return
 }
+
+func GetVirtualMachineSizesFromMoc() (vMSizes string, err error) {
```
madhanrm, line 457: Lets not convert this to string here.
Lets return an array of the new type.

Lets rename the method to GetVirtualMachineSizesDetailed

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/spf13/viper v1.7.1
 	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013 // indirect
 	google.golang.org/grpc v1.27.1
+	github.com/microsoft/moc-pkg v0.10.13-alpha.1
```
madhanrm, line 14: we cannot use moc-pkg. This is a private repo and moc-sdk-for-go is a public repo

services/compute/vmSizes.go
acraghav: addressed first 2 comments, 3rd comment I took liberty to name it as GetVirtualMachineSizeValues, let me know if you still don't like, I will rename it back

go.mod
acraghav: done. moved the constants to moc-common

madhanrm (approved):

madhanrm:
can you revendor this in moccli and update the cmdlet

https://github.com/microsoft/moccli/blob/master/cmd/mocctl/compute/virtualmachine/list_sizes/virtualmachine_list_sizes.go

PR #96, opened 2022-02-23
adding hypervgeneration param

hvedati:

PR #97, opened 2022-03-02
Image download/provision support

pathuang-sia:
Passing down urn and cloudinitdatasource from moc-operator to cloud agent

### DIFF
--- a/services/compute/compute.go
+++ b/services/compute/compute.go
@@ -761,6 +761,8 @@ type GalleryImageProperties struct {
 	OsType OperatingSystemTypes `json:"osType,omitempty"`
 	// OsState - This property allows the user to specify whether the virtual machines created under this image are 'Generalized' or 'Specialized'. Possible values include: 'Generalized', 'Specialized'
 	OsState OperatingSystemStateTypes `json:"osState,omitempty"`
+	// HyperVGeneration - The hypervisor generation of the Virtual Machine. Applicable to OS disks only. Possible values include: 'V1', 'V2'
+	// HyperVGeneration HyperVGeneration `json:"hyperVGeneration,omitempty"`
 	// EndOfLifeDate - The end of life date of the gallery Image Definition. This property can be used for decommissioning purposes. This property is updatable.
 	EndOfLifeDate *date.Time                       `json:"endOfLifeDate,omitempty"`
 	Identifier    *GalleryImageIdentifier          `json:"identifier,omitempty"`
@@ -777,8 +779,6 @@ type GalleryImageProperties struct {
 	SourceType common.ImageSource `json:"sourceType,omitempty"`
 	// CloudInitDataSource - READ-ONLY; The cloud init data source to be used with the image. [NoCloud, Azure]. Default Value – NoCloud. For marketplace images it will be Azure.
 	CloudInitDataSource common.CloudInitDataSource `json:"cloudInitDataSource,omitempty"`
-	// DownlaodStatus - Current download status of the image
-	DownloadStatus common.DownloadStatus `json:"downloadStatus,omitempty"`
 }
 
 // GalleryImage specifies information about the gallery Image Definition that you want to create or update.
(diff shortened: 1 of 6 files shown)


services/compute/compute.go
```diff
@@ -666,6 +666,8 @@ type GalleryImageIdentifier struct {
 	Offer *string `json:"offer,omitempty"`
 	// Sku - The name of the gallery Image Definition SKU.
 	Sku *string `json:"sku,omitempty"`
+	// Version - Specifies the version of the platform image or marketplace image used to create the virtual machine.
+	Version *string `json:"version,omitempty"`
```
madhanrm, line 670: This property doesnt exist in Azure sdk. Why are we introducing it here
https://github.com/Azure/azure-sdk-for-go/blob/main/services/compute/mgmt/2021-11-01/compute/models.go#L7900

services/compute/compute.go
```diff
@@ -753,8 +755,6 @@ type GalleryImageProperties struct {
 	OsType OperatingSystemTypes `json:"osType,omitempty"`
 	// OsState - This property allows the user to specify whether the virtual machines created under this image are 'Generalized' or 'Specialized'. Possible values include: 'Generalized', 'Specialized'
 	OsState OperatingSystemStateTypes `json:"osState,omitempty"`
-	// HyperVGeneration - The hypervisor generation of the Virtual Machine. Applicable to OS disks only. Possible values include: 'V1', 'V2'
```
madhanrm, line 756: Any reason to remove this comment?

services/compute/compute.go
```diff
@@ -769,6 +769,10 @@ type GalleryImageProperties struct {
 	ContainerName *string `json:"containername,omitempty"`
 	//Type of source of gal image (sfs/http/local)
 	SourceType common.ImageSource `json:"sourceType,omitempty"`
+	// CloudInitDataSource - READ-ONLY; The cloud init data source to be used with the image. [NoCloud, Azure]. Default Value – NoCloud. For marketplace images it will be Azure.
+	CloudInitDataSource common.CloudInitDataSource `json:"cloudInitDataSource,omitempty"`
+	// DownlaodStatus - Current download status of the image
+	DownloadStatus common.DownloadStatus `json:"downloadStatus,omitempty"`
```
madhanrm, line 775: Try to add DownloadStatus into Status key value map
It doesnt make sense here

services/compute/compute.go
madhanrm: The Version is specific to http download from market place. 
Check how we do this using SFSImageProperties. Either use httpImageProperties or define a new struct and use it the same way as SfsImageProperties and pass it along as a string.

Try not to introduce params that would be specific to a type

madhanrm: And Please document that changes in the design md

services/compute/compute.go
pathuang-sia: The value will be added in this PR: https://github.com/microsoft/moc-sdk-for-go/pull/96/files. I'll remove this change.

madhanrm: still see this change. Can you remove this change?

services/compute/compute.go
madhanrm: Should this be DownloadStatus or upload status - since we are uploading the image to MOC

manoj-kadam: It's tricky. In doing that it's also downloading the image from Azure Marketplace. In the ARM API it's proposed as DownloadStatus.

pathuang-sia: I'll move downloadstatus under status but will need to modify moc/pkg/status to pass it down. I'll link the pr here once ready.

madhanrm: @pathuang-sia  - I think the ask was different and Manoj answerd it. Thanks.

pathuang-sia: https://github.com/microsoft/moc/pull/116

madhanrm (approved):

PR #98, opened 2022-03-08
add keyRotationFrequencyInSeconds to Key properties

flowerbunny1988:
This PR added a new field to allow configuration key rotation frequency.

Design doc: https://github.com/microsoft/wssdagent/pull/618

Testing:
New test cases added to WssdTesting and being tested locally that covers:
1. If key rotation frequency is set to -1, then the AES key does not rotate
2. If key rotation frequency is set to positive value, then the AES key rotates

OS tracking task for ASZ: [Task 37702743](https://microsoft.visualstudio.com/OS/_workitems/edit/37702743): ASZ R1: Implement the option to not do key rotation

### DIFF
--- a/services/security/keyvault/key/key.go
+++ b/services/security/keyvault/key/key.go
@@ -37,9 +37,10 @@ func getKey(sec *wssdcloudsecurity.Key, vaultName string) (keyvault.Key, error) 
 		Version: &sec.Status.Version.Number,
 		Value:   &value,
 		KeyProperties: &keyvault.KeyProperties{
-			Statuses: status.GetStatuses(sec.GetStatus()),
-			KeyType:  getKeyType(sec.Type),
-			KeySize:  keysize,
+			Statuses:                      status.GetStatuses(sec.GetStatus()),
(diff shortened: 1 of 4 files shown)


services/security/keyvault/key/key.go
```diff
@@ -67,13 +68,14 @@ func getWssdKey(name string, sec *keyvault.Key,
...
-		KeyOps:    []wssdcloudcommon.KeyOperation{},
-		Status:    status.InitStatus(),
+		Name:                          name,
+		VaultName:                     vaultName,
+		GroupName:                     groupName,
+		Type:                          getMOCKeyType(sec.KeyType),
+		Size:                          keysize,
+		KeyOps:                        []wssdcloudcommon.KeyOperation{},
+		Status:                        status.InitStatus(),
+		KeyRotationFrequencyInSeconds: *sec.KeyRotationFrequencyInSeconds,
```
raghavendra-nataraj, line 78: You may want a null check here

madhanrm (approved):
Good to merge

PR #99, opened 2022-03-21
GalleryImage SizeInBytes support with moc version update

pathuang-sia:
Update moc package version to latest for adding support for passing downloaded image size from bottom to top; the value will be surfaced to the operator level.

### DIFF
--- a/services/compute/compute.go
+++ b/services/compute/compute.go
@@ -779,8 +779,6 @@ type GalleryImageProperties struct {
 	SourceType common.ImageSource `json:"sourceType,omitempty"`
 	// CloudInitDataSource - READ-ONLY; The cloud init data source to be used with the image. [NoCloud, Azure]. Default Value – NoCloud. For marketplace images it will be Azure.
 	CloudInitDataSource common.CloudInitDataSource `json:"cloudInitDataSource,omitempty"`
-	// SizeInBytes - This property indicates the size of the gallery image in Bytes. (Int64 format)
-	SizeInBytes *int64 `json:"sizeInBytes,omitempty"`
 }
 
 // GalleryImage specifies information about the gallery Image Definition that you want to create or update.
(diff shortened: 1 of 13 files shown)


manoj-kadam (approved):

services/compute/compute.go
```diff
@@ -779,6 +779,8 @@ type GalleryImageProperties struct {
 	SourceType common.ImageSource `json:"sourceType,omitempty"`
 	// CloudInitDataSource - READ-ONLY; The cloud init data source to be used with the image. [NoCloud, Azure]. Default Value – NoCloud. For marketplace images it will be Azure.
 	CloudInitDataSource common.CloudInitDataSource `json:"cloudInitDataSource,omitempty"`
+	// SizeInBytes - This property indicates the size of the gallery image in Bytes. (Int64 format)
+	SizeInBytes *int64 `json:"sizeInBytes,omitempty"`
```
madhanrm, line 783: do we have this property in Azure sdk?

madhanrm, line 783: Do we have this property in azure sdk?

pathuang-sia: It will be equivalent to this: https://github.com/Azure/azure-sdk-for-go/blob/bea25d2e2710632d5814b1a01a0b19ff10561ac6/services/compute/mgmt/2021-11-01/compute/models.go#L9061

Kept it as bytes due to the need of converting it to megabytes at the operator level.

madhanrm: But GalleryImageProperties doesnt have this and we are trying to pull in the GalleryImageVersion property into this, and basically are changing the definition.
For now, can you put this in status, until we have a requirement to bring in GalleryImageVersion

pathuang-sia: In that case, I'll move the sizeInBytes under cloudagent galleryimage to downloadstatus and modify moc/pkg/status to set the sizeInBytes as well when setting downloadStatus: https://github.com/microsoft/moc/pull/119

madhanrm (approved):

manoj-kadam (approved):

PR #101, opened 2022-03-23
Initial code change for Moc sdk wrapper for cpp

flowerbunny1988:
This change creates a wrapper dll that cpp code can call into Moc APIs. This is the initial change. More function calls will be added.
The scenario is for ASZ TVM 22H2 release key management flow. A cpp written component needs to call into Moc APIs to perform key operations.

Validation:
1. Has integrated this change with the cpp component (outside Moc repo) that all function calls work as expected. There will be separate PR in the cpp repo with integration and testing code.
2. Build succeeds in current repo.

Tracking work item by ASZ:
[Task 37933487](https://microsoft.visualstudio.com/OS/_workitems/edit/37933487): ASZ R1: Create wrapper component so that the GSP request sent from native component can be forwarded to MocStack


wrapper/cpp/main.go
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache v2.0 License.
+
+//
+// This file contains wrapper function calls that c++ component
+// can leverage to call into MocStack
+//
+
+package main
```
madhanrm, line 9: Is there a way we could have some directory structure as go services, so we can expand adding more apis.

flowerbunny1988: I dont know how to implement file structures. What I read is all exported functions needs to be coded up in the go file where package is main and there is main() function. Do you know how to implement it? I think what you are suggesting is to have, for example, EncryptData() under wrapper\cpp\security\keyvault\key\ folder.

If your team can help provide guidance how to do that, I can open a task to track this. At this point we are very close to code complete, hence I prefer to get this change in to unblock our team from integrating with attestation agent.

madhanrm: At this point we dont have details. Lets go with this and refactor it later.

flowerbunny1988: Thanks Madhan. All checks are passing now, is it okay to merge?

wrapper/cpp/main.go
```diff
@@ -0,0 +1,142 @@
...
+	var value string
+	value = C.GoString(input)
+
+	parameters := &keyvault.KeyOperationsParameters{
+		Value:     &value,
+		Algorithm: keyvault.A256KW,
+	}
+
+	response, err := keyClient.Encrypt(ctx, C.GoString(groupName), C.GoString(keyvaultName), C.GoString(keyName), parameters)
+    if err != nil {
```
penglimsft, line 41: nit: i think this if is mis-aligned

wrapper/cpp/main.go
```diff
@@ -0,0 +1,142 @@
...
+    // check the length and return 1 (means key exists) if there is more than one key
+    if keys != nil && len(*keys) > 0 {
+        return  1
+    }
+
+    return 0
+}
+
+//export KeyvaultKeyCreateOrUpdateAESKey
+func KeyvaultKeyCreateOrUpdateAESKey(serverName *C.char, groupName *C.char, keyvaultName *C.char, keyName *C.char, timeoutInSeconds C.int) *C.char {
```
madhanrm, line 99: KeyvaultKeyCreateOrUpdate

wrapper/cpp/main.go
```diff
@@ -0,0 +1,142 @@
...
+
+    var kvConfig *keyvault.Key
+    kvConfig = &keyvault.Key{}
+
+    var keyNameString string
+    keyNameString = C.GoString(keyName)
+    kvConfig.Name = &keyNameString
+    kvConfig.KeyProperties = &keyvault.KeyProperties{}
+
+    kvConfig.KeyType = keyvault.JSONWebKeyType("AES")
```
madhanrm, line 113: can you make this a param?

madhanrm (approved):

madhanrm:
if we can add a test to validate if the dll load works , it would be great

PR #105, opened 2022-04-05
Add cloudinitdatasource support to cloudagent level vhd

pathuang-sia:
Add cloudinitdatasource support to cloudagent level vhd and update moc version

### DIFF
--- a/services/storage/virtualharddisk/virtualharddisk.go
+++ b/services/storage/virtualharddisk/virtualharddisk.go
@@ -62,7 +62,11 @@ func getWssdVirtualHardDisk(c *storage.VirtualHardDisk, groupName, containerName
 		} else {
 			wssdvhd.DiskFileFormat = common.DiskFileFormat_DiskFileFormatVHDX
 		}
-
+		if &c.CloudInitDataSource != nil {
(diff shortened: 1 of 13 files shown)


services/storage/virtualharddisk/virtualharddisk.go
```diff
@@ -62,7 +62,11 @@ func getWssdVirtualHardDisk(c *storage.VirtualHardDisk, groupName, containerName
 		} else {
 			wssdvhd.DiskFileFormat = common.DiskFileFormat_DiskFileFormatVHDX
 		}
-
+		if &c.CloudInitDataSource != nil {
```
manoj-kadam, line 65: This condition will always be true, isn't it? You are comparing address of a variable against nil pointer. Why do we need this check?

pathuang-sia: considering backwards compatibility since it's possible it might not be specified from moc-sdk-for-go when mocctl creates a storage.VirtualHardDisk directly

manoj-kadam (approved):

madhanrm (approved):

PR #107, opened 2022-04-21
Add security login to cpp wrapper

penglimsft:
Add a new cpp wrapper API to allow for Security Login using a login config file.


madhanrm:
Good to merge

PR #110, opened 2022-04-28
Updating SDK to support key import/export APIs

niranjanshnd:
We need to implement these APIs to enable disaster recovery and inter-cluster VM migration. Design documentation can be found at: https://github.com/microsoft/wssdagent/blob/user/nishinde/keyimportexport/doc/cloudagent-key-import-export.md

### DIFF
--- a/services/security/keyvault/key/client.go
+++ b/services/security/keyvault/key/client.go
@@ -5,6 +5,7 @@ package key
 
 import (
 	"context"
+
 	"github.com/microsoft/moc-sdk-for-go/services/security"
 	"github.com/microsoft/moc-sdk-for-go/services/security/keyvault"
 	"github.com/microsoft/moc/pkg/auth"
@@ -19,6 +20,7 @@ type Service interface {
 	Decrypt(context.Context, string, string, string, *keyvault.KeyOperationsParameters) (*keyvault.KeyOperationResult, error)
 	WrapKey(context.Context, string, string, string, *keyvault.KeyOperationsParameters) (*keyvault.KeyOperationResult, error)
 	UnwrapKey(context.Context, string, string, string, *keyvault.KeyOperationsParameters) (*keyvault.KeyOperationResult, error)
+	ImportKey(context.Context, string, string, string, *keyvault.Key) (*keyvault.Key, error)
 }
 
 // Client structure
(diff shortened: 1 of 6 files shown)


services/security/keyvault/keyvault.go
```diff
@@ -200,6 +202,39 @@ type Key struct {
 	*KeyProperties `json:"properties,omitempty"`
 }
 
+// KeyWrappingAlgorithm enumerates the values for private key wrapping info for key import export operations.
```
madhanrm, line 205: https://github.com/Azure/azure-sdk-for-go/blob/main/services/keyvault/v7.0/keyvault/models.go

am looking at azure sdk for reference. Would prefer to have same APIs

services/security/keyvault/key/client.go
```diff
@@ -48,6 +51,18 @@ func (c *KeyClient) CreateOrUpdate(ctx context.Context, group, vaultName, name s
 	return c.internal.CreateOrUpdate(ctx, group, vaultName, name, param)
 }
```
madhanrm, line 53: Would like to bring in the same APIs with same params as much as  possible,
Please refer https://github.com/Azure/azure-sdk-for-go/blob/98ef3b076ed6af3696bda23cad806d69e482e989/services/keyvault/v7.0/keyvault/client.go#L129

niranjanshnd: The key import and export scenario in MOC resembles the AKVs BYOK (Already public: https://docs.microsoft.com/en-us/azure/key-vault/keys/byok-specification) and Secure Key Release (In beta/private testing: https://microsoft.sharepoint.com/:w:/t/AzureKeyVault45/EbWaKoYIYlpOn2TKwQ91ijgBCh-m-ln5yOnLSeXdiM6w8A). The currently public backup API of AKV is not intended to provide key in a usable format outside of the AKV, so it does not address the disaster recovery scenario critical to ASZ. I have tried to match the parameters and crypto algorithm of the new APIs to resemble the BYOK and SKR. You can read more about the design choices in my design document in the PR (https://github.com/microsoft/wssdagent/pull/628).

services/security/keyvault/key/wssd.go
```diff
@@ -86,6 +89,185 @@ func (c *client) CreateOrUpdate(ctx context.Context, group, vaultName, name stri
...
+// Import
+func (c *client) ImportKey(ctx context.Context, group, vaultName, name string, param *keyvault.Key) (*keyvault.Key, error) {
+	err := c.validate(ctx, group, vaultName, name, param)
+	if err != nil {
+		return nil, err
+	}
+	if param.KeySize == nil {
+		return nil, errors.Wrapf(errors.InvalidInput, "Invalid KeySize - Missing")
+	}
+	request, err := getKeyRequest(wssdcloudcommon.Operation_IMPORT, group, vaultName, name, param)
```
madhanrm, line 187: This function is okay, but instead of using Operation_IMPORT, can you reuse Operation_POST

raghavendra-nataraj (approved):

raghavendra-nataraj (approved):

PR #113, opened 2022-06-24
Show owner node name for VM, attached node name for VHD

in3xes:


services/compute/virtualmachine/virtualmachine.go
```diff
@@ -414,6 +414,7 @@ func (c *client) getVirtualMachine(vm *wssdcloudcompute.VirtualMachine, group st
 			VmType:                  vmtype,
 			DisableHighAvailability: &vm.DisableHighAvailability,
 			Host:                    c.getVirtualMachineHostDescription(vm),
+			NodeName:                &vm.NodeName,
```
madhanrm, line 417: This is hidden for a reason. Why do we need this?
We cannot expose physical nodes to clients that creates vms

PR #118, opened 2022-09-30
Adding CKM_RSA_AES Import/Export support for AES Keys

SethBe:
Adding required changes to support returning the key wrap information for AES keys.


penglimsft (approved):

services/security/keyvault/key/key.go
```diff
@@ -51,8 +51,16 @@ func getKey(sec *wssdcloudsecurity.Key, vaultName string, getCustomValue func(*w
 			}
 			value = string(pem.EncodeToMemory(&pubBlk))
 		}
+	case wssdcloudcommon.JsonWebKeyType_AES:
```
raghavendra-nataraj, line 54: Nit: Can we add a default with error specifying invalid type ?

SethBe: I dont think we want to do that.
The current default, is to return empty string and not fail.  This is true if you call this for a normal AES key without the getcustomvalue callback. I dont think we should change this behavior.

We could do this if the callback is provided, but i think its better to keep consistent (empty string = failure)

raghavendra-nataraj (approved):

PR #125, opened 2023-01-11
Add new parameters to enable identity auto rotation

ganeshamohana:


services/security/identity/identity.go
```diff
@@ -44,6 +44,8 @@ func getIdentity(id *wssdcloudsecurity.Identity) *security.Identity {
 			CloudPort:     &id.CloudPort,
 			CloudAuthPort: &id.CloudAuthPort,
 		},
+		AutoRotate:    id.AutoRotate,
```
raghavendra-nataraj, line 47: Can we add a check that the loginpath is passed of autorotate is enabled? This will help up return early

raghavendra-nataraj (approved):

PR #126, opened 2023-01-21
Add identity tags in response for Get operation

ganeshamohana:


services/security/identity/identity.go
```diff
@@ -27,6 +27,14 @@ func getIdentity(id *wssdcloudsecurity.Identity) *security.Identity {
 		clitype = auth.LoadBalancer
 	}
 
+	tags := make(map[string]*string)
```
raghavendra-nataraj, line 30: Let's try to use the same patter 
https://github.com/microsoft/moc-sdk-for-go/blob/main/services/compute/virtualmachine/wssd.go#L290

raghavendra-nataraj (changes requested):

services/security/identity/identity.go
ganeshamohana: fixed

raghavendra-nataraj (approved):

PR #129, opened 2023-02-09
fix valid connection

raghavendra-nataraj:


pkg/client/client.go
```diff
@@ -90,7 +90,7 @@ func isValidConnections(conn *grpc.ClientConn) bool {
 	if conn.GetState() == connectivity.TransientFailure {
```
apluchik, line 90: nit: we can probably change this to a switch statement

apluchik (approved):
Looks good.

apluchik (approved):

PR #130, opened 2023-02-13
added grpc interceptor and singleton changes

Kaivalya1997:


pkg/constant/constant.go
```diff
@@ -12,3 +13,24 @@ const (
...
+type ClientOpts struct {
+	NoExitOnConnFailure bool
+}
+
+var clientOptsSingleton *ClientOpts
+var once sync.Once
+
+//If GetClientOpts called before SetClientOpts, it will return a default ClientOpts
+func GetClientOpts() *ClientOpts {
+	once.Do(func() {
```
raghavendra-nataraj, line 26: Wouldn't this cause an issue if set is called before get. The clientOptsSingleton would be nil

raghavendra-nataraj: if once.Do(f) is called multiple times, only the first call will invoke f, even if f has a different value in each invocation. A new instance of Once is required for each function to execute.

raghavendra-nataraj: https://pkg.go.dev/sync#Once.Do

Kaivalya1997: So the expected behaviour is to call Set before Get, which the developer needs to ensure. In case if Set is not called and Get is invoked first, the default value of the NoExitOnConnFailure will be set to false as once.Do will execute once and set it once finally. In case Get is called before, the value will not be nil. As the NoExitOnConnFailure flag is set at the beginning of the mocctl process, I am ensuring that for mocctl the Set is being called before any Get operations occur

raghavendra-nataraj: Cant we just check if clientOptsSingleton is not nil and return default if nil?

Kaivalya1997: We can do that as well, but we would need to take care of thread safety explicitly. I used once.Do because it is inherently thread safe. Do you want me to implement the nil checking logic?

msftpgayam:
@Kaivalya1997 do we still want to pick up this PR? If not, can we close it?

PR #132, opened 2023-02-15
RBAC Provider Operation Support

ArulvelMuralidharan:
RBAC Provider Operation Support


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/Azure/go-autorest/autorest v0.9.0
 	github.com/Azure/go-autorest/autorest/date v0.2.0
 	github.com/google/uuid v1.3.0
-	github.com/microsoft/moc v0.11.0-alpha.2
+	github.com/microsoft/moc v0.10.29-alpha.3
```
madhanrm, line 10: Are we going back on the versioning?

ArulvelMuralidharan: Ragha said he was using this tag for his own testing. I think v0.10.29-alpha.3 is the latest tag

madhanrm: Please create something like v0.11.0-alpha.3 and use that as latest.
v0.10 is old

ArulvelMuralidharan: Resolved

services/security/security.go
```diff
@@ -163,12 +163,64 @@ const (
 type Operation string
 
 const (
-	ReadAccess   Operation = "read"
-	WriteAccess  Operation = "write"
-	DeleteAccess Operation = "delete"
-	AllAccess    Operation = "all"
+	OBSOLETE_ReadAccess   Operation = "read"
```
raghavendra-nataraj, line 166: Dowe know if they are being used outside this repo?

ArulvelMuralidharan: Don't believe so. Only place I would think is moccli but I checked and they don't there.

raghavendra-nataraj (approved):

PR #133, opened 2023-02-20
Added service to add threshold loglevel

abhadoriya26:


services/admin/logging/internal/wssd.go
```diff
@@ -66,8 +67,34 @@ func (c *client) GetLogFile(ctx context.Context, location, filename string) erro
...
+
 func getLoggingRequest(location string) *wssdadmin.LogRequest {
 	return &wssdadmin.LogRequest{
 		Location: location,
 	}
 }
+
+func setLogLevelRequest(setloglevel uint32) *wssdadmin.SetRequest {
+	return &wssdadmin.SetRequest{
+		Settracelevel: setloglevel,
```
in3xes, line 98: nit: let's keep the naming consistent. Either use trace level or log level. Not both.

PR #135, opened 2023-02-27
protobuf cve fix

raghavendra-nataraj:


services/cloud/cluster/wssd.go
```diff
@@ -126,7 +126,7 @@ func (c *client) Unload(ctx context.Context, location, name string) error {
 	return err
 }
 
-///////////////////////////
+// /////////////////////////
```
madhanrm, line 129: Is this change expected ?

madhanrm (approved):

services/cloud/cluster/wssd.go
raghavendra-nataraj: I think vs code auto format must have done this

PR #143, opened 2023-03-07
Update LoginFilepath checks

ruinan-liu:
Updated 2 checks to ensure that if `AutoRotate` is true and `LoginFilePath` is not `nil`, then we check
1. `LoginFilePath` is not an empty string ""
2. `LoginFilePath` is an absolute file path

### DIFF
--- a/services/security/identity/identity.go
+++ b/services/security/identity/identity.go
@@ -4,6 +4,7 @@
 package identity
 
 import (
+	"fmt"
 	"path/filepath"
 
 	"github.com/microsoft/moc-sdk-for-go/services/security"
@@ -59,6 +60,7 @@ func getWssdIdentity(id *security.Identity) (*wssdcloudsecurity.Identity, error)
 	}
 
 	if id.LoginFilePath != nil && !(filepath.IsAbs(*id.LoginFilePath)) {
+		fmt.Println(*id.LoginFilePath)
 		return nil, errors.Wrapf(errors.InvalidInput, "Identity Loginfile must be absolute filepath")
 	}
 	wssdidentity := &wssdcloudsecurity.Identity{
--- a/services/security/identity/identity.go
+++ b/services/security/identity/identity.go
@@ -4,7 +4,6 @@
 package identity
 
 import (
-	"fmt"
 	"path/filepath"
 
 	"github.com/microsoft/moc-sdk-for-go/services/security"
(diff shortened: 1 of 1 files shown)


services/security/identity/identity.go
```diff
@@ -113,6 +110,17 @@ func getWssdIdentity(id *security.Identity) (*wssdcloudsecurity.Identity, error)
 
 	wssdidentity.ClientType = clitype
 	wssdidentity.AutoRotate = id.AutoRotate
+
+	// In case auto rotate in enabled, the login file path can not be empty or relatively
+	if id.AutoRotate == true && id.LoginFilePath != nil {
+		if *id.LoginFilePath == "" {
+			return nil, errors.Wrapf(errors.InvalidInput, "Identity Loginfile path cannot be empty")
```
raghavendra-nataraj, line 117: This should not return a error. Why are we returning a error here?

ruinan-liu: Are we not writing to the location of LoginFilePath if we have enabled token auto rotation?

raghavendra-nataraj: No, when autorotate is enabled, the file path can be empty or can be absolute path. It is not necessary for the path to be there for autorotate

ruinan-liu: Understood, Done

services/security/identity/identity.go
```diff
@@ -113,6 +110,14 @@ func getWssdIdentity(id *security.Identity) (*wssdcloudsecurity.Identity, error)
 
 	wssdidentity.ClientType = clitype
 	wssdidentity.AutoRotate = id.AutoRotate
+
+	// In case auto rotate in enabled, the login file path can not be relative
+	if id.AutoRotate == true && id.LoginFilePath != nil {
+		if !(filepath.IsAbs(*id.LoginFilePath)) {
```
raghavendra-nataraj, line 116: This will throw an error when the LoginFilePath is empty, leading to the same issue

ruinan-liu: Thanks, sorry that I missed this case. I have added a condition so that it will only check whether a path is absolute if
1. The auto rotate flag is enabled
2. The login file path is not nil
3. Login file path is not empty

raghavendra-nataraj (approved):

services/security/identity/identity.go
raghavendra-nataraj: Can we write a few units test so that we don't regress in future

PR #144, opened 2023-03-08
Adding Unit test for function getWssdidentity in Identity.go

ruinan-liu:
Manually run `make test` below is the result
image


raghavendra-nataraj:
Please can you update the makefile to run this test?

Makefile
```diff
@@ -39,3 +39,6 @@ test:
 
 format:
 	gofmt -s -w pkg/ services/ 
+
+test-security:
```
raghavendra-nataraj, line 43: Nit: lets call it unittest

raghavendra-nataraj: We can add more tests tomorrow that are not security related

.pipelines/build.yaml
```diff
@@ -32,6 +32,7 @@ jobs:
   - script: |
       sudo apt-get install gcc-mingw-w64
       mkdir manifest
+      make test-security
```
raghavendra-nataraj, line 35: we can add the test-security target to all. This way we don't have to add a separate command

raghavendra-nataraj (approved):

PR #145, opened 2023-03-09
Propagate isPrimary flag from moccli to wssdCloudAgent

gsicilianomsft:
Propagate the Primary flag for an ipConfig from the moccli to wssdCloudAgent


services/network/networkinterface/networkinterface.go
```diff
@@ -126,6 +126,9 @@ func getWssdNetworkInterfaceIPConfig(ipConfig *network.InterfaceIPConfiguration)
 	if ipConfig.Gateway != nil {
 		wssdipconfig.Gateway = *ipConfig.Gateway
 	}
+	if ipConfig.Primary != nil {
```
nijosmsft, line 129: please update getNetworkIpConfig also so that Primary shows up correctly in get calls.

nijosmsft (changes requested):
update getNetworkIpConfig

in3xes (approved):

PR #146, opened 2023-03-10
Bump down go bump for moc-sdk

madhanrm:


apluchik (dismissed):

pkg/client/client.go
```diff
@@ -45,9 +43,6 @@ func isDebugMode() error {
 	if debugEnv == "on" {
 		return nil
 	}
-	if viper.GetBool("Debug") {
```
apluchik, line 48: do we use this flag in our testing at all?

madhanrm: we shouldnt test insecure mode anymore

apluchik (approved):

PR #172, opened 2023-08-10
Added instrumentation to CppWrapper

ryanlampe-msft:
This PR added instrumentation to the CppWrapper

Design doc: https://github.com/microsoft/wssdagent/pull/789

Testing:
See Design doc.


wrapper/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+		return "", ""
+	}
+	client, err := version.NewVersionClient(serverName, authorizer)
+	if err != nil {
+		return "", ""
+	}
+	mocVersion, agentVersion, err := client.GetVersion(context.Background())
+	if err != nil {
+		return "", ""
+	}
```
ncampuzano, line 114: A suggestion, as "" means error on getting Moc version; what about we change it to something like "Error on new authorizer from environment" , "Error getting new version client", "Error getting version". So once we saw in the logs we can determinate the exact error. As now we will see a "" and knows it was an error, but no which.

wrapper/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+			etw.StringField("WrapperTelemetryVersion", WrapperTelemetryVersion),
+			etw.StringField("MocVersion", eventTelemetry.MocVersion),
+			etw.StringField("MocAgentVersion", eventTelemetry.MocAgentVersion),
+		))
+	if err != nil {
+		return
+	}
+}
+
+func getMocVersion(serverName string) (string, string) {
```
SethBe, line 102: I dont see where this is used?
Note i think this is internal to the wrapper. You can return err as well

ncampuzano: Do you mean `getMocVersion`? That is used on `telemetry.go` line 53: `mocVersion, mocAgentVersion = getMocVersion(serverName)`

wrapper/telemetry.go
ryanlampe-msft: This is a good thought. "" does mean that we don't know the moc version. However, from a DI perspective, we don't want to clutter the version name with error information. 
I will update the getMocVersion function to additionally & separately return an error. The error won't be used in the context of this PR because there is nothing to do with the error, but if others leverage the function, they may want to.

SethBe (dismissed):
:shipit:

in3xes (approved):

randy-danielson (dismissed):

raghavendra-nataraj (dismissed):

raghavendra-nataraj (approved):

SethBe (approved):
:shipit:

PR #177, opened 2023-09-21
Change encryption/decryption algorithm to A256CBC

ganeshamohana:


SethBe (dismissed):
:shipit:

SethBe:
LOoks good to me from a Mocwrapper perspective. Please share the bins when ready and i can do some smoke testing.

services/security/keyvault/key/key.go
```diff
@@ -257,6 +257,8 @@ func getMOCAlgorithm(algo keyvault.JSONWebKeyEncryptionAlgorithm) (wssdcloudcomm
 		return wssdcloudcommon.Algorithm_RSAOAEP256, nil
 	case keyvault.A256KW:
```
raghavendra-nataraj, line 258: Are we still supporting this?

ganeshamohana: No, we'll reject the request for this.

raghavendra-nataraj: Is this used by someother api call, if not we can remove it

raghavendra-nataraj (approved):

SethBe (approved):
:shipit:

PR #180, opened 2023-09-23
User/dhwanishah/adding proxy support

djshah19:


services/compute/virtualmachine/virtualmachine.go
```diff
@@ -418,6 +421,32 @@ func (c *client) getWssdVirtualMachineGuestAgentConfiguration(s *compute.GuestAg
 	return gac, nil
 }
 
+func (c *client) getWssdVirtualMachineHttpProxyConfiguration(httpProxyConfig *compute.HttpProxyConfiguration) *wssdcloudcompute.HttpProxyConfiguration {
```
raghavendra-nataraj, line 424: Can we add unit tests validating this functions.

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -418,6 +425,59 @@ func (c *client) getWssdVirtualMachineGuestAgentConfiguration(s *compute.GuestAg
...
+		proxyConfiguration.TrustedCa = *proxyConfig.TrustedCa
+	}
+
+	if proxyConfig.HttpProxy != nil {
+		if opType == wssdcloudproto.Operation_POST {
+			parsedUrl, err := validations.ValidateProxyURL(*proxyConfig.HttpProxy)
+			if err != nil {
+				return nil, errors.Wrapf(errors.InvalidInput, err.Error())
+			}
+			err = validations.TestProxyUrlConnection(parsedUrl, *proxyConfig.TrustedCa, "")
```
raghavendra-nataraj, line 451: I think we should not have URL connection check from moc-sdk. we may not have line of sight.

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -401,6 +403,11 @@ func (c *client) getWssdVirtualMachineOSConfiguration(s *compute.OSProfile) (*ws
 	if s.CustomData != nil {
 		osconfig.CustomData = *s.CustomData
 	}
+
+	osconfig.ProxyConfiguration, err = c.getWssdVirtualMachineProxyConfiguration(s.ProxyConfiguration, opType)
+	if err != nil {
```
raghavendra-nataraj, line 408: How are we enforcing the update of proxy settings is blocked?

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -4,17 +4,19 @@
...
+	"github.com/microsoft/moc/pkg/validations"
 
 	"github.com/microsoft/moc/pkg/status"
 	wssdcloudcompute "github.com/microsoft/moc/rpc/cloudagent/compute"
+	wssdcloudproto "github.com/microsoft/moc/rpc/common"
 	wssdcommon "github.com/microsoft/moc/rpc/common"
 )
 
-func (c *client) getWssdVirtualMachine(vm *compute.VirtualMachine, group string) (*wssdcloudcompute.VirtualMachine, error) {
+func (c *client) getWssdVirtualMachine(vm *compute.VirtualMachine, group string, opType wssdcloudproto.Operation) (*wssdcloudcompute.VirtualMachine, error) {
```
raghavendra-nataraj, line 19: What operation are we restricting this from? Why are we adding the check for POST operation

djshah19: Below validations should be done only for post operation

services/compute/virtualmachine/virtualmachine.go
djshah19: why?

services/compute/virtualmachine/virtualmachine.go
djshah19: Arc VM operators, Arc VM cli, and moccli doesnt support update.

raghavendra-nataraj: How can we enforce the same in the moc-sdk?

services/compute/virtualmachine/virtualmachine.go
raghavendra-nataraj: Moc-sdk can be called from any host, it need not have access to the proxy. Only the VM needs to have access, we should be validating from SDK

services/compute/virtualmachine/virtualmachine.go
raghavendra-nataraj: What are the implications if we pass the proxy value for other operations? I thought the other requests should be noop. Correct me if I am wrong

djshah19: reason why I added post operation check is because for other operations we get redacted values which fails the validations

services/compute/virtualmachine/virtualmachine.go
raghavendra-nataraj: Because the restriction is from Moc(cloudagent/nodeagent) that we dont support moc, I think we should restrict this is moc-sdk or cloudagent

djshah19: I dont see how we can restrict update in moc sdk

raghavendra-nataraj (approved):

services/compute/virtualmachine/wssd.go
```diff
@@ -305,3 +310,22 @@ func getComputeTags(tags *wssdcloudproto.Tags) map[string]*string {
 func getWssdTags(tags map[string]*string) *wssdcloudproto.Tags {
 	return prototags.MapToProto(tags)
 }
+
+func (c *client) virtualMachineValidations(opType wssdcloudproto.Operation, vmss *compute.VirtualMachine) error {
```
msftpgayam, line 314: We need to check for `OsProfile` nil value, right?

I don't think there's anything stopping from nil osconfig value here.

djshah19: yes, good catch

msftpgayam: nit: we won't come here if vmss is nil, I think we can avoid that for readability.

djshah19: done!

msftpgayam (approved):

PR #183, opened 2023-10-18
Add OsVersion to Tags

vishal-kadam:
1. Add OsVersion to Tags of Node object.
2. Code changes as part of [this](https://msazure.visualstudio.com/DefaultCollection/One/_workitems/edit/25064743) feature.
3. osVersion will be needed in the Operators to check if the Node is pre23H2 or not.
4. 
MicrosoftTeams-image (11)


ggaurav10 (approved):
LGTM

PR #187, opened 2023-10-27
fix debug flag

raghavendra-nataraj:


madhanrm (approved):

pkg/client/client.go
```diff
@@ -43,6 +44,9 @@ func isDebugMode() error {
 	if debugEnv == "on" {
 		return nil
 	}
+	if viper.GetBool("Debug") {
```
madhanrm, line 47: Why are we using this flag?

madhanrm: if we use it in moccli, then it should remain there. We cannot extend its use into sdk

PR #196, opened 2023-12-19
Add support for Network Security Groups

vlappenbusch:


services/network/networksecuritygroup/wssd.go
```diff
@@ -0,0 +1,327 @@
...
+		return fmt.Errorf("Network Security Group [%s] not found", name)
+	}
+
+	request, err := c.getNetworkSecurityGroupRequest(wssdcloudcommon.Operation_DELETE, group, name, &(*nsgs)[0])
+	if err != nil {
+		return err
+	}
+	_, err = c.NetworkSecurityGroupAgentClient.Invoke(ctx, request)
+
+	if err != nil {
```
msftpgayam, line 94: nit: We are returning the anyway, no need for nil check

services/network/networksecuritygroup/wssd.go
```diff
@@ -0,0 +1,327 @@
...
+	if networkNSG.Name == nil {
+		return nil, errors.Wrapf(errors.InvalidConfiguration, "Missing Name for Network Security Group")
+	}
+
+	wssdCloudNSG = &wssdcloudnetwork.NetworkSecurityGroup{
+		Name:      *networkNSG.Name,
+		GroupName: group,
+	}
+
+	if networkNSG.Location != nil {
```
msftpgayam, line 162: Do we need both group and location?

services/network/networksecuritygroup/wssd.go
```diff
@@ -0,0 +1,327 @@
...
+			wssdCloudNSGRule.Protocol = wssdcloudcommon.Protocol_All
+		} else if strings.EqualFold(string(rule.Protocol), string(network.SecurityRuleProtocolTCP)) {
+			wssdCloudNSGRule.Protocol = wssdcloudcommon.Protocol_Tcp
+		} else if strings.EqualFold(string(rule.Protocol), string(network.SecurityRuleProtocolUDP)) {
+			wssdCloudNSGRule.Protocol = wssdcloudcommon.Protocol_Udp
+		} else {
+			return nil, errors.Wrapf(errors.InvalidInput, "Unknown Protocol %s specified", rule.Protocol)
+		}
+
+		if rule.SourceAddressPrefix != nil {
```
msftpgayam, line 213: Since SDK exposes `SourceAddressPrefixes` we will need to handle that case as well.

services/network/networksecuritygroup/wssd.go
```diff
@@ -0,0 +1,327 @@
...
+		if rule.SecurityRulePropertiesFormat == nil {
+			continue
+		}
+
+		wssdCloudNSGRule := &wssdcloudnetwork.NetworkSecurityGroupRule{}
+
+		if rule.Name == nil {
+			return nil, errors.Wrapf(errors.InvalidInput, "Network Security Rule name not specified")
+		}
+		wssdCloudNSGRule.Name = *rule.Name
```
msftpgayam, line 197: Check for duplicate names

services/network/network.go
```diff
@@ -16,6 +16,10 @@ const (
 	TransportProtocolTCP TransportProtocol = "Tcp"
 	// TransportProtocolUDP
 	TransportProtocolUDP TransportProtocol = "Udp"
+	// TransportProtocolICMPv4
+	TransportProtocolICMPv4 TransportProtocol = "Icmpv4"
```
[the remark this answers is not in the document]
vlappenbusch: They're not, it looks like. Removed

services/network/networksecuritygroup/wssd.go
vlappenbusch: handled, along with the other plurals!

services/network/networksecuritygroup/wssd.go
vlappenbusch: checked in createOrUpdate!

services/network/network.go
```diff
@@ -796,7 +800,7 @@ type SecurityRulePropertiesFormat struct {
 	// Access - The network traffic is allowed or denied. Possible values include: 'SecurityRuleAccessAllow', 'SecurityRuleAccessDeny'
 	Access SecurityRuleAccess `json:"access,omitempty"`
 	// Priority - The priority of the rule. The value can be between 100 and 4096. The priority number must be unique for each rule in the collection. The lower the priority number, the higher the priority of the rule.
```
sbgms, line 802: >100 and 4096 [](http://example.com/codeflow?start=66&length=12)

0 to 65536

services/network/networksecuritygroup/wssd.go
```diff
@@ -0,0 +1,360 @@
...
+			wssdCloudNSGRule.Description = *rule.Description
+		}
+
+		if strings.EqualFold(string(rule.Protocol), string(network.SecurityRuleProtocolAsterisk)) {
+			wssdCloudNSGRule.Protocol = wssdcloudcommon.Protocol_All
+		} else if strings.EqualFold(string(rule.Protocol), string(network.SecurityRuleProtocolTCP)) {
+			wssdCloudNSGRule.Protocol = wssdcloudcommon.Protocol_Tcp
+		} else if strings.EqualFold(string(rule.Protocol), string(network.SecurityRuleProtocolUDP)) {
+			wssdCloudNSGRule.Protocol = wssdcloudcommon.Protocol_Udp
+		} else {
```
sbgms, line 218: Also need to support Http

services/network/networksecuritygroup/wssd.go
```diff
@@ -0,0 +1,360 @@
...
+
+		if rule.SourcePortRange != nil {
+			wssdCloudNSGRule.SourcePortRange = *rule.SourcePortRange
+		} else if rule.SourcePortRanges != nil {
+			concatRule := ""
+			for _, prefix := range *rule.SourcePortRanges {
+				concatRule += prefix
+			}
+			wssdCloudNSGRule.SourcePortRange = concatRule
+		}
```
sbgms, line 250: Should we parse any malformed inputs ?

services/network/networksecuritygroup/wssd.go
```diff
@@ -0,0 +1,360 @@
...
+			wssdCloudNSGRule.Direction = wssdcloudnetwork.Direction_Outbound
+		} else {
+			return nil, errors.Wrapf(errors.InvalidInput, "Unknown Direction %s specified", rule.Access)
+		}
+
+		if rule.Priority != nil && isValidPriority(*rule.Priority) {
+			wssdCloudNSGRule.Priority = uint32(*rule.Priority)
+		} else {
+			wssdCloudNSGRule.Priority = 4096 // TODO: what should be the default value?
+		}
```
sbgms, line 282: Priority is incorrect. You can have from (100 to 65500)

services/network/network.go
```diff
@@ -891,8 +895,6 @@ type PrivateEndpoint struct {
 type InterfacePropertiesFormat struct {
 	// VirtualMachine - READ-ONLY; The reference of a virtual machine.
 	VirtualMachine *SubResource `json:"virtualMachine,omitempty"`
```
sbgms, line 897: Does removing NetworkSecurityGroup break any older clients ?  @msftpgayam thoughts ?

services/network/networksecuritygroup/wssd.go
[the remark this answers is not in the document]
vlappenbusch: These don't have matching inputs from azure - they only specify ICMP. Should that match to one of these, or potentially either of these in different contexts?

services/network/network.go
vlappenbusch: In Azure it's 100 to 4096. I don't think we're failing operations if it's out of this range before network controller though

services/network/networksecuritygroup/wssd.go
vlappenbusch: I think the logic for it is complicated enough that I'd want to just let NC handle it, but you're right, we could add in the future.

services/network/network.go
msftpgayam: Since it's not being used right now, I don't think it will break the older client. It should be ok to remove a field in the sdk.

msftpgayam (approved):

msftpgayam (approved):

sbgms (dismissed):

nijosmsft (approved):
:shipit:

PR #220, opened 2024-03-06
Add NSG support to logical network

vlappenbusch:
Wait until NSG PR is merged


services/network/network.go
```diff
@@ -24,6 +24,14 @@ type SubResource struct {
 	ID *string `json:"id,omitempty"`
 }
 
+// GroupSubResource referes to another subresource, and the group that subresource belongs to
+type GroupSubResource struct {
+	// ID - Resource ID.
+	ID *string `json:"id,omitempty"`
```
nijosmsft, line 30: >ID [](http://example.com/codeflow?start=1&length=2)

in azure a resource id is supposed to be fully qualified, for eg; /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/resourceType/{resourceName}

as far as i understand, we try to keep moc-sdk similar to azure. that means we dont need a new groupsubresource. 

any reason to diverge from this? @msftpgayam

vlappenbusch: discussed offline - I believe the plan is to keep this as is

services/network/network.go
```diff
@@ -24,6 +24,14 @@ type SubResource struct {
 	ID *string `json:"id,omitempty"`
 }
 
+// GroupSubResource referes to another subresource, and the group that subresource belongs to
+type GroupSubResource struct {
```
mjethwa-msft, line 28: can you just briefly explain here why are we referencing the resource differently here?

mjethwa-msft (dismissed):

services/network/network.go
vlappenbusch: Sorry didn't see this earlier! We did have this previously because we wanted to reference an NSG in an LNET, but we coulnd't just use the NSG name because NSG was parented to group while LNET was parented to location, so we couldn't use the LNET's group. That said, because of this problem we decided to change to have NSG parented at the location level - so this was removed again

mjethwa-msft (approved):

services/network/virtualnetwork/virtualnetwork.go
```diff
@@ -136,7 +136,13 @@ func getWssdNetworkSubnets(subnets *[]network.Subnet) (wssdsubnets []*wssdcloudn
 		}
 
 		if subnet.NetworkSecurityGroup != nil {
-			wssdsubnet.NetworkSecurityGroup = *subnet.NetworkSecurityGroup.ID
+			wssdsubnet.NetworkSecurityGroup = &wssdcommonproto.ResourceReference{
+				Name: *subnet.NetworkSecurityGroup.ID,
+			}
+
+			if location != nil {
+				wssdsubnet.NetworkSecurityGroup.Location = *location
```
nijosmsft, line 144: >wssdsubnet.NetworkSecurityGroup.Location = *location [](http://example.com/codeflow?start=4&length=52)

qq, this is now optional right?

vlappenbusch: yes! I had it not optional for a second, but fixed it. The wssd definition doesn't really have a way to mark a field as optional, but I'm only consistently filling in the name

msftpgayam, line 144: We should error out if `location` is nil, right? Or `nil` location is being handled in the cloud gent code?

vlappenbusch: removed!

PR #221, opened 2024-03-13
API for Resize with GPUs

dhavalpopat:
Design Doc: https://microsoft.sharepoint.com/:w:/t/AzureEdgeTeam/EYQGzpguvPRDmSOejqFOyyABiRzfzas0glXutZlBhvW14w?e=aW0HA9

### DIFF
--- a/services/compute/compute.go
+++ b/services/compute/compute.go
@@ -217,13 +217,28 @@ type DynamicMemoryConfiguration struct {
 	TargetMemoryBuffer *uint32 `json:"targetmemorybuffer,omitempty"`
 }
 
+type Assignment string
+
+// possible values of gpu assignment
+const (
+	GpuDDA     Assignment = "GpuDDA"
+	GpuP       Assignment = "GpuP"
+	GpuPV      Assignment = "GpuPV"
+	GpuDefault Assignment = "GpuDefault"
+)
+
+type VirtualMachineGPU struct {
+	Assignment      *Assignment `json:"assignment,omitempty"`
+	PartitionSizeMB *uint64     `json:"partitionSizeMB,omitempty"`
+	Name            *string     `json:"name,omitempty"`
+}
+
 type HardwareProfile struct {
-	VMSize     VirtualMachineSizeTypes   `json:"vmsize,omitempty"`
-	CustomSize *VirtualMachineCustomSize `json:"customsize,omitempty"`
+	VMSize             VirtualMachineSizeTypes   `json:"vmsize,omitempty"`
+	VirtualMachineGPUs []*VirtualMachineGPU      `json:"virtualMachineGPUs,omitempty"`
+	CustomSize         *VirtualMachineCustomSize `json:"customsize,omitempty"`
 	// DynamicMemoryConfig - Specifies the dynamic memory configuration for a VM, dynamic memory will be enabled if this field is present.
 	DynamicMemoryConfig *DynamicMemoryConfiguration `json:"dynamicmemoryconfig,omitempty"`
-	// VirtualMachineGPUs - Specifies list of GPUs for VM
-	VirtualMachineGPUs []*common.VirtualMachineGPU `json:"virtualmachinegpus,omitempty"`
 }
 
 // NetworkInterfaceReferenceProperties describes a network interface reference properties.
--- a/services/compute/compute.go
+++ b/services/compute/compute.go
@@ -203,9 +203,8 @@ type OSProfile struct {
 
 // VirtualMachineCustomSize Specifies cpu/memory information for custom VMSize types.
 type VirtualMachineCustomSize struct {
-	CpuCount *int32        `json:"cpucount,omitempty"`
-	MemoryMB *int32        `json:"memorymb,omitempty"`
-	GpuList  []*common.Gpu `json:"gpulist,omitempty"`
+	CpuCount *int32 `json:"cpucount,omitempty"`
+	MemoryMB *int32 `json:"memorymb,omitempty"`
 }
 
 // DynamicMemoryConfiguration Specifies the dynamic memory configuration for a VM.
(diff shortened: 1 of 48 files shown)


services/compute/virtualmachine/client.go
```diff
@@ -137,7 +137,7 @@ func (c *VirtualMachineClient) Update(ctx context.Context, group string, vmName
 }
 
 // Resize the Virtual Machine
-func (c *VirtualMachineClient) Resize(ctx context.Context, group string, vmName string, newSize compute.VirtualMachineSizeTypes, newCustomSize *compute.VirtualMachineCustomSize) (err error) {
+func (c *VirtualMachineClient) Resize(ctx context.Context, group string, vmName string, hardware *compute.HardwareProfile) (err error) {
```
manoj-kadam, line 140: Won't this be a breaking change?

manoj-kadam: Why not just add one more param to pass in list of GPUs instead of passing in h/w profile. I guess changing dynamic memory configuration is not supported

services/compute/compute.go
```diff
@@ -222,6 +222,8 @@ type HardwareProfile struct {
 	CustomSize *VirtualMachineCustomSize `json:"customsize,omitempty"`
 	// DynamicMemoryConfig - Specifies the dynamic memory configuration for a VM, dynamic memory will be enabled if this field is present.
 	DynamicMemoryConfig *DynamicMemoryConfiguration `json:"dynamicmemoryconfig,omitempty"`
+	// VirtualMachineGPUs - Specifies list of GPUs for VM
+	VirtualMachineGPUs []*common.VirtualMachineGPU `json:"virtualmachinegpus,omitempty"`
```
dhavalpopat, line 226: Don't use common, create new object here

services/compute/virtualmachine/client.go
dhavalpopat: Oh understood, yes will just add a new param for gpu

services/compute/compute.go
dhavalpopat: Done

services/compute/virtualmachine/client.go
dhavalpopat: Done

blairurish (dismissed):

services/compute/virtualmachine/client.go
```diff
@@ -139,6 +139,11 @@ func (c *VirtualMachineClient) Update(ctx context.Context, group string, vmName
 
 // Resize the Virtual Machine
 func (c *VirtualMachineClient) Resize(ctx context.Context, group string, vmName string, newSize compute.VirtualMachineSizeTypes, newCustomSize *compute.VirtualMachineCustomSize) (err error) {
+	return c.ResizeEx(ctx, group, vmName, newSize, newCustomSize, nil)
```
raghavendra-nataraj, line 142: Is there a reason we are passing nil for the vmGPUs

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -209,11 +210,15 @@ func (c *client) getWssdVirtualMachineHardwareConfiguration(vm *compute.VirtualM
...
 				}
-				vmGPU := &wssdcommon.VirtualMachineGPU{
-					Assignment:      assignment,
-					PartitionSizeMB: *gpu.PartitionSizeMB,
-					Name:            *gpu.Name,
+				vmGPU.Assignment = assignment
+				if gpu.Name != nil {
+					vmGPU.Name = *gpu.Name
+				}
+				if gpu.PartitionSizeMB != nil {
```
raghavendra-nataraj, line 220: Is PartitionSizeMB a optional parameter?

dhavalpopat: It's not an optional parameter. Should I remove this check?

raghavendra-nataraj: If it is not optional , should we not be checking if both the parameters are set?

manoj-kadam: Yes, we should fail if assignment type is GPU-P and partition size is not specified.

dhavalpopat: Apologies for the confusion. For Arc VMs, we need to provide partition size from portal and CLI because that's how we calculate the number of partitions or number of DDA devices. But for other platforms, it can be empty too depending on how they define and implement in their operators/CLI.
 
But from moc perspective, since we are getting vmGPUs array in hardware profile, there is no need for partition size to be mandatory.
 
So, from moc perspective, name and partition size parameters are optional. AssignmentType is required.

raghavendra-nataraj: So what would it mean if we dont set the partition size?

services/compute/virtualmachinescaleset/virtualmachinescaleset.go
```diff
@@ -403,10 +432,48 @@ func (c *client) getWssdVirtualMachineScaleSetHardwareConfiguration(vmp *compute
...
+					// if partition size is not specified, set it to 0
+					*gpu.PartitionSizeMB = 0
+				}
+				if gpu.Name == nil {
+					// if name is not specified, set it to empty string
+					*gpu.Name = ""
+				}
+				vmGPU := &wssdcommon.VirtualMachineGPU{
+					Assignment:      assignment,
+					PartitionSizeMB: *gpu.PartitionSizeMB,
```
manoj-kadam, line 466: Shouldn't we have nil check here too?

manoj-kadam (approved):

blairurish (approved):

services/compute/virtualmachine/virtualmachine.go
dhavalpopat: For GPU-P homogeneous env, there is only one partition size available, so that will be used.

For GPU-P heterogeneous env, we will return an error. 

For GPU-DDA, partition size is not required/ not applicable.

dhavalpopat:
Closing in favor of - https://github.com/microsoft/moc-sdk-for-go/pull/247

From moc team:

> I saw the same thing for a previous PR where I was making a PR from a fork of the main repo - was never able to figure out what the issue was, but when I tried building off a branch in the main repo it worked

Created a new PR from a branch in microsoft original repo.

PR #223, opened 2024-03-21
Fix json property name

mjethwa-msft:
Fixes JSON property name in `LogicalSubnetPropertiesFormat`.  

It is a one-line change to match the json property name with the struct property name.

### Public *bool `json:"public,omitempty"``


nijosmsft (approved):
:shipit:

vlappenbusch (approved):

PR #227, opened 2024-04-04
add vm discovery for cloudagent

katl-msft:


services/compute/compute.go
```diff
@@ -1160,3 +1160,12 @@ type ProxyConfiguration struct {
 	// Alternative CA cert to use for connecting to proxy server
 	TrustedCa *string `json:"trustedca,omitempty"`
 }
+
+type VirtualMachineDiscovery struct {
```
mduppre, line 1164: As we discussed yesterday, I think we don't need this struct after all as we're just using the `VirtualMachine` struct

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -457,12 +457,12 @@ func (c *client) getVirtualMachine(vm *wssdcloudcompute.VirtualMachine, group st
 		vmtype = compute.StackedControlPlane
 	}
 	return &compute.VirtualMachine{
-		Name:                     &vm.Name,
-		ID:                       &vm.Id,
-		Tags:                     getComputeTags(vm.GetTags()),
+		Name: &vm.Name,
```
mduppre, line 460: Would you mind reverting the whitespace changes? Obviously very low priority, just will make our overall PR easier to read

mduppre (approved):
Looks good :-)

services/compute/virtualmachine/virtualmachine.go
neerajdixit-msft2: I think these are done automatically by `make fmt` according to go rules so better keep it as is.

mduppre: Makes sense!

services/compute/compute.go
neerajdixit-msft2: @mduppre I think we will need this structure if we want to change the API function signature.

mduppre: yeah, you're right, I wrote this comment before we chatted about the API function signature. So let's leave it and I'll look into converting the objects

PR #233, opened 2024-05-07
Fix nsg concatenation

vlappenbusch:
NSG fields that use multiple inputs, such as SourceAddressPrefixes, should be mapped to a single field by concatenating and comma separating the array into one field. The comma separation was missed, making this SDK work not as intended for inputs with multiple values in the array.

This changes to use comma separation for array input fields, as well as the standardized string concatenation function through golang.

### DIFF
--- a/services/network/networksecuritygroup/wssd.go
+++ b/services/network/networksecuritygroup/wssd.go
@@ -271,7 +271,7 @@ func getWssdNetworkSecurityGroupRules(securityRules *[]network.SecurityRule, isD
 			return nil, errors.Wrapf(errors.InvalidInput, "Unknown Protocol %s specified", rule.Protocol)
 		}
 
-		if rule.SourceAddressPrefix != nil {
+		if rule.SourceAddressPrefix != nil && len(*rule.SourceAddressPrefix) > 0 {
 			wssdCloudNSGRule.SourceAddressPrefix = *rule.SourceAddressPrefix
 		} else if rule.SourceAddressPrefixes != nil {
 			concatRule := ""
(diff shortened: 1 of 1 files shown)


nijosmsft:
hi @vlappenbusch , is this still required? or did the operators changed their format?

vlappenbusch:
Yes, this should still be merged, I wasn't able to get testing put together for it yet but it should be fixed! I believe the operators have a workaround for now

services/network/networksecuritygroup/wssd.go
```diff
@@ -271,44 +271,28 @@ func getWssdNetworkSecurityGroupRules(securityRules *[]network.SecurityRule, isD
...
-		if rule.SourceAddressPrefix != nil {
+		if rule.SourceAddressPrefix != nil && len(*rule.SourceAddressPrefix) > 0 {
 			wssdCloudNSGRule.SourceAddressPrefix = *rule.SourceAddressPrefix
 		} else if rule.SourceAddressPrefixes != nil {
-			concatRule := ""
-			for _, prefix := range *rule.SourceAddressPrefixes {
-				concatRule += prefix
-			}
-			wssdCloudNSGRule.SourceAddressPrefix = concatRule
+			wssdCloudNSGRule.SourceAddressPrefix = strings.Join(*rule.SourceAddressPrefixes, ",")
```
nijosmsft, line 277: >strings.Join(*rule.SourceAddressPrefixes, ",") [](http://example.com/codeflow?start=42&length=46)

i remember you mentioned operators did some workarounds because of this bug, will those workarounds continue to work once we fix it? or do we need operators to change?

vlappenbusch: Yes, I'll reach out to them to remove those fixes! I believe their fix was to concatenate the arrays themselves and to pass an array with one element that is a concatenated string, so that would still work. I can make sure to notify Arka and whoever else needs to know when this is merged.

nijosmsft: ok, as this is low pri fix, let's hold on and dont merge until we are 100% sure. i would recommend sharing a private with them and testing. for context, our asz private preview is going to happen in 2408 we need to be extra careful with anything we merge.

mjethwa-msft (dismissed):

PR #234, opened 2024-05-20
moc sdk for go change for hybrid aks gpu

wenjlu-lgy:


services/compute/virtualmachine/virtualmachine.go
```diff
@@ -201,6 +201,7 @@ func (c *client) getWssdVirtualMachineHardwareConfiguration(vm *compute.VirtualM
 		VMSize:                     sizeType,
 		CustomSize:                 customSize,
 		DynamicMemoryConfiguration: dynMemConfig,
+		VirtualMachineGPUs:         vm.HardwareProfile.VirtualMachineGPUs,
```
msftpgayam, line 204: We have to add this field in `getVirtualmachinehardwareprofile()` as well, right?

Could you please build moccli using these changes and make sure moccli is printing the GPUs?

services/compute/compute.go
```diff
@@ -217,9 +217,26 @@ type DynamicMemoryConfiguration struct {
 	TargetMemoryBuffer *uint32 `json:"targetmemorybuffer,omitempty"`
 }
 
+type Assignment string
+
+// possible values of gpu assignment
+const (
```
msftpgayam, line 223: Why are we having to redefine the constants here? are they not available in the moc repo?

wenjlu-lgy: Redefine is used to translate int32 in moc to string. Supposing it will make sdk more user friendly and avoid exposing underlying moc details to customers. It's following the "SecurityTypes" definition (also translate int32 in moc to string in moc-sdk-for-go).

msftpgayam (approved):

msftpgayam (approved):

PR #245, opened 2024-06-20
Adding SecurityEncryptionType

shayan-msft:
Changes:

- Addition of SecurityEncryptionType and parent structs
- Changes to pass this type through the stack

### DIFF
--- a/services/compute/virtualmachine/virtualmachine.go
+++ b/services/compute/virtualmachine/virtualmachine.go
@@ -598,6 +598,8 @@ func (c *client) getVirtualMachineStorageProfileOsDisk(d *wssdcloudcompute.Disk)
 			switch d.ManagedDisk.SecurityProfile.SecurityEncryptionType {
 			case wssdcommon.SecurityEncryptionTypes_NonPersistedTPM:
 				securityEncryptionType = compute.NonPersistedTPM
+			default:
+				securityEncryptionType = ""
 			}
 			managedDisk.SecurityProfile = &compute.VMDiskSecurityProfile{
 				SecurityEncryptionType: securityEncryptionType,
--- a/services/compute/virtualmachine/virtualmachine.go
+++ b/services/compute/virtualmachine/virtualmachine.go
@@ -144,14 +144,36 @@ func (c *client) getWssdVirtualMachineStorageConfigurationImageReference(s *comp
 	return *s.Name, nil
 }
 func (c *client) getWssdVirtualMachineStorageConfigurationOsDisk(s *compute.OSDisk) (*wssdcloudcompute.Disk, error) {
-	if s.Vhd == nil {
-		return nil, errors.Wrapf(errors.InvalidInput, "Vhd Configuration is missing in OSDisk")
+	if s.ManagedDisk == nil {
+		if s.Vhd == nil {
+			return nil, errors.Wrapf(errors.InvalidInput, "Vhd Configuration is missing in OSDisk")
+		}
+		if s.Vhd.URI == nil {
+			return nil, errors.Wrapf(errors.InvalidInput, "Vhd URI Configuration is missing in OSDisk")
+		}
 	}
-	if s.Vhd.URI == nil {
-		return nil, errors.Wrapf(errors.InvalidInput, "Vhd URI Configuration is missing in OSDisk")
+
+	diskName := ""
+	if s.Vhd != nil && s.Vhd.URI != nil {
+		diskName = *s.Vhd.URI
+	}
+	var managedDisk *wssdcommon.VirtualMachineManagedDiskParameters
+	if s.ManagedDisk != nil {
+		managedDisk = &wssdcommon.VirtualMachineManagedDiskParameters{}
+		if s.ManagedDisk.SecurityProfile != nil {
+			securityEncryptionType := wssdcommon.SecurityEncryptionTypes_SecurityEncryptionNone
+			switch s.ManagedDisk.SecurityProfile.SecurityEncryptionType {
+			case compute.NonPersistedTPM:
+				securityEncryptionType = wssdcommon.SecurityEncryptionTypes_NonPersistedTPM
+			}
+			managedDisk.SecurityProfile = &wssdcommon.VMDiskSecurityProfile{
+				SecurityEncryptionType: securityEncryptionType,
+			}
+		}
 	}
 	return &wssdcloudcompute.Disk{
-		Diskname: *s.Vhd.URI,
+		Diskname:    diskName,
+		ManagedDisk: managedDisk,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


danmamsft (approved):

msftpgayam (approved):

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -144,14 +144,36 @@ func (c *client) getWssdVirtualMachineStorageConfigurationImageReference(s *comp
...
+
+	diskName := ""
+	if s.Vhd != nil && s.Vhd.URI != nil {
+		diskName = *s.Vhd.URI
+	}
+	var managedDisk *wssdcommon.VirtualMachineManagedDiskParameters
+	if s.ManagedDisk != nil {
+		managedDisk = &wssdcommon.VirtualMachineManagedDiskParameters{}
+		if s.ManagedDisk.SecurityProfile != nil {
+			securityEncryptionType := wssdcommon.SecurityEncryptionTypes_SecurityEncryptionNone
```
SethBe, line 164: >securityEncryptionType := wssdcommon.SecurityEncryptionTypes_SecurityEncryptionNone [](http://example.com/codeflow?start=3&length=83)

should we do this outside the if statement, so that its always set?

shayan-msft: we could, but we only construct the VMDisksecurityProfile inside the if, so we wouldn't use the securityEncryptionType outside the if block

SethBe: nvm not sure that make sense.

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -144,14 +144,36 @@ func (c *client) getWssdVirtualMachineStorageConfigurationImageReference(s *comp
...
+			switch s.ManagedDisk.SecurityProfile.SecurityEncryptionType {
+			case compute.NonPersistedTPM:
+				securityEncryptionType = wssdcommon.SecurityEncryptionTypes_NonPersistedTPM
+			}
```
SethBe, line 168: consider adding a default case to none

msftpgayam (approved):

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -144,14 +144,38 @@ func (c *client) getWssdVirtualMachineStorageConfigurationImageReference(s *comp
...
+			return nil, errors.Wrapf(errors.InvalidInput, "Vhd URI Configuration is missing in OSDisk")
+		}
 	}
-	if s.Vhd.URI == nil {
-		return nil, errors.Wrapf(errors.InvalidInput, "Vhd URI Configuration is missing in OSDisk")
+
+	diskName := ""
+	if s.Vhd != nil && s.Vhd.URI != nil {
+		diskName = *s.Vhd.URI
+	}
```
msftpgayam, line 159: we should fail if any of them are `nil`, right?

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -144,14 +144,38 @@ func (c *client) getWssdVirtualMachineStorageConfigurationImageReference(s *comp
...
+	var managedDisk *wssdcommon.VirtualMachineManagedDiskParameters
+	if s.ManagedDisk != nil {
+		managedDisk = &wssdcommon.VirtualMachineManagedDiskParameters{}
+		if s.ManagedDisk.SecurityProfile != nil {
+			securityEncryptionType := wssdcommon.SecurityEncryptionTypes_SecurityEncryptionNone
```
msftpgayam, line 164: also, why set `None` if this is handled in `default`?

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -564,8 +588,23 @@ func (c *client) getVirtualMachineStorageProfileOsDisk(d *wssdcloudcompute.Disk)
 	if d == nil {
 		return &compute.OSDisk{}
 	}
+	var managedDisk *compute.VirtualMachineManagedDiskParameters
+	if d.ManagedDisk != nil {
+		managedDisk = &compute.VirtualMachineManagedDiskParameters{}
+		if d.ManagedDisk.SecurityProfile != nil {
+			var securityEncryptionType compute.SecurityEncryptionTypes
```
msftpgayam, line 595: Should we set to type `None` by default?

msftpgayam (changes requested):

msftpgayam (commented):
commented on the wrong PR. Github doesn't allow me to delete a comment

services/compute/virtualmachine/virtualmachine.go
shayan-msft: it's just for clarity, it will default to None without the switch anyways.
If you think it looks better without I'll remove it

services/compute/virtualmachine/virtualmachine.go
shayan-msft: if ManagedDisk is set, then these can be nil, since we're using ManagedDisk to set the SecurityProfile, but we still want to be able to use a gallery image (so we would set only ManagedDisk and leave vhd as nil)

services/compute/virtualmachine/virtualmachine.go
shayan-msft: sure, I'll set it to none by default for consistency, although the type already defaults to none

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -144,14 +144,38 @@ func (c *client) getWssdVirtualMachineStorageConfigurationImageReference(s *comp
 	return *s.Name, nil
 }
 func (c *client) getWssdVirtualMachineStorageConfigurationOsDisk(s *compute.OSDisk) (*wssdcloudcompute.Disk, error) {
-	if s.Vhd == nil {
-		return nil, errors.Wrapf(errors.InvalidInput, "Vhd Configuration is missing in OSDisk")
+	if s.ManagedDisk == nil {
```
raghavendra-nataraj, line 147: Why are we adding the VHD check inside the managedisk?

shayan-msft: I added a comment, hopefully that makes it clearer.
If ManagedDisk == nil it's the normal VM creation from OsDisk and we want to error if Vhd is invalid.
Otherwise, we allow Vhd to be invalid since we may be creating VM from imagereference

msftpgayam (approved):

msftpgayam:
Good to merge. 

Could you rebase the PR.

raghavendra-nataraj (dismissed):

msftpgayam (approved):

danmamsft (approved):

PR #246, opened 2024-06-26
Add other resource precheck functions

aeweston-msft:
# Changes introduced by this PR
- Add Precheck helper functions that call the corresponding new precheck RPC introduced in https://github.com/microsoft/moc/pull/269 for the following services:
- Identity (security)
- Availability set (compute)
- Gallery image (compute)
- Virtual machine image (compute)
- Load balancer (network)
- Logical network (network)
- MAC pool (network)
- Network interface (network)
- Network security group (network)
- VIP pool (network)
- Virtual network (network)
- Container (storage)

# Test plan
Have sanity checked the new precheck API for the storage container resource including cloudagent and moc-sdk-for-go changes. Since the Precheck RPC definition is common except for the resource types, I think this is a sufficient minimum of testing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/Azure/go-autorest/autorest v0.11.29
 	github.com/Azure/go-autorest/autorest/date v0.3.0
 	github.com/google/uuid v1.4.0
-	github.com/microsoft/moc v0.19.2
+	github.com/microsoft/moc v0.19.3
 	google.golang.org/grpc v1.59.0
 	k8s.io/klog v1.0.0
 )
@@ -24,7 +24,6 @@ replace (
 	github.com/Azure/go-autorest v11.1.2+incompatible => github.com/Azure/go-autorest/autorest v0.10.0
 	github.com/dgrijalva/jwt-go => github.com/golang-jwt/jwt/v4 v4.1.0
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
-	github.com/microsoft/moc => github.com/microsoft/moc v0.19.2-0.20240621211241-c2bd73cfc238
 	github.com/miekg/dns => github.com/miekg/dns v1.1.25
 	github.com/nats-io/nkeys => github.com/nats-io/nkeys v0.4.6
 	golang.org/x/net => golang.org/x/net v0.0.0-20220822230855-b0a4917ee28c
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ replace (
 	github.com/Azure/go-autorest v11.1.2+incompatible => github.com/Azure/go-autorest/autorest v0.10.0
 	github.com/dgrijalva/jwt-go => github.com/golang-jwt/jwt/v4 v4.1.0
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
+	github.com/microsoft/moc => github.com/microsoft/moc v0.19.2-0.20240621211241-c2bd73cfc238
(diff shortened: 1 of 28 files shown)


go.mod
```diff
@@ -24,6 +24,7 @@ replace (
 	github.com/Azure/go-autorest v11.1.2+incompatible => github.com/Azure/go-autorest/autorest v0.10.0
 	github.com/dgrijalva/jwt-go => github.com/golang-jwt/jwt/v4 v4.1.0
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
+	github.com/microsoft/moc => github.com/microsoft/moc v0.19.2-0.20240621211241-c2bd73cfc238
```
aeweston-msft, line 27: Needs to be updated once MOC PR goes in

services/network/logicalnetwork/wssd.go
```diff
@@ -79,6 +79,52 @@ func (c *client) Delete(ctx context.Context, location, name string) error {
...
+func getLogicalNetworkPrecheckRequest(location string, logicalNetworks []*network.LogicalNetwork) (*wssdcloudnetwork.LogicalNetworkPrecheckRequest, error) {
+	request := &wssdcloudnetwork.LogicalNetworkPrecheckRequest{}
+
+	protoLogicalNetworks := make([]*wssdcloudnetwork.LogicalNetwork, 0, len(logicalNetworks))
+
+	for _, logicalNetwork := range logicalNetworks {
+		// can logical network ever be nil here? what would be the meaning of that?
+		if logicalNetwork != nil {
+
+			// TODO (aweston): double check this
```
aeweston-msft, line 103: This was one thing I wasn't sure about - all of the other getWssd* functions take location/group as separate parameters, but not this one. @mjethwa-msft can you give any context on the intent of that logic in getWssdLogicalNetwork/getLogicalNetworkRequest?

raghavendra-nataraj: Logical network is a location based entitity

aeweston-msft: Sorry for the confusion, the part I was unsure about is that all the other resources have location or group as a separate param to the equivalent conversion function, whereas here it is not a param, but expected to be filled out in the SDK struct. I was not sure if there was any reason that was done, since the top level functions still have location as a separate parameter. The getWssdLogicalNetwork function will fail if it is nil, so it seems odd to have two different ways to pass it. Mostly just wanted to check if there was something I was missing here, and whether it is always expected to be filled out in the SDK struct itself.

To give an example:

func getWssdLogicalNetwork(c *network.LogicalNetwork) 
vs. virtual machine image where group is a separate parameter
func getWssdVirtualMachineImage(c *compute.VirtualMachineImage, **groupName string)**

raghavendra-nataraj: Ok, sorry got the question wrong. @mjethwa-msft Was there a reason for this? 

Today Lnet are created from Moc_operator only, so may be they always pass the location in the struct. I thing it was just an assumption made

raghavendra-nataraj (approved):

madhanrm (approved):

raghavendra-nataraj (approved):

services/network/logicalnetwork/wssd.go
aeweston-msft: Going to go ahead and merge as is and will follow up separately. I think what I have implemented is safe and reasonable, just would be good to clarify the invariant here (if any) so we could simplify the logic.

PR #249, opened 2024-07-12
VirtualHardDisk Download and GalleryImage Create From OS Disk

damanMSFT:


services/storage/storage.go
```diff
@@ -40,6 +40,10 @@ type VirtualHardDiskProperties struct {
 	CloudInitDataSource common.CloudInitDataSource `json:"cloudInitDataSource,omitempty"`
 	// Container name
 	ContainerName *string `json:"containername,omitempty"`
+	// Source path for the VHD. Only supports url.
```
msftpgayam, line 43: Can we keep it consistent with Azure SDK? Here's the pointer to the Disk struct in Azure SDK: https://github.com/Azure/azure-sdk-for-go/blob/745d4a5bb9c400c77d32dbd4046331f24e3cb2c9/sdk/resourcemanager/compute/armcompute/models.go#L1453

services/storage/virtualharddisk/client.go
```diff
@@ -74,3 +77,17 @@ func (c *VirtualHardDiskClient) Resize(ctx context.Context, group, container, na
 func (c *VirtualHardDiskClient) Precheck(ctx context.Context, group, container string, vhds []*storage.VirtualHardDisk) (bool, error) {
 	return c.internal.Precheck(ctx, group, container, vhds)
 }
+
+func (c *VirtualHardDiskClient) DownloadVhdFromHttp(ctx context.Context, group, container, name string, storage *storage.VirtualHardDisk, azHttpImg *compute.AzureGalleryImageProperties) (*storage.VirtualHardDisk, error) {
```
msftpgayam, line 81: Can we call it `UploadDiskFromHttp` to keep it consistent with the GalleryImage API.

go.mod
```diff
@@ -24,6 +24,7 @@ replace (
 	github.com/Azure/go-autorest v11.1.2+incompatible => github.com/Azure/go-autorest/autorest v0.10.0
 	github.com/dgrijalva/jwt-go => github.com/golang-jwt/jwt/v4 v4.1.0
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
+	github.com/microsoft/moc => ../moc
```
msftpgayam, line 27: Could you fix the path here to point to your change in moc repo on github. This doc describes how to do it.

https://msazure.visualstudio.com/msk8s/_wiki/wikis/msk8s.wiki/646685/Testing-dependent-PRs

services/storage/storage.go
damanMSFT: I am not sure how else I should add these properties. I don't think azure has a concept of downloading disk from somewhere like we are planning to do here.

services/storage/virtualharddisk/client.go
damanMSFT: I think we should keep it DownloadVhdFromHttp as we are actually performing a download here and not Upload. Also we should rename them to download as well in GalleryImage API to avoid confusion in a seperate PR. I don't think writing upload is the correct way here. Let me know your thoughts.

services/compute/compute.go
```diff
@@ -852,6 +852,8 @@ type GalleryImageProperties struct {
 	SourceType common.ImageSource `json:"sourceType,omitempty"`
 	// CloudInitDataSource - READ-ONLY; The cloud init data source to be used with the image. [NoCloud, Azure]. Default Value – NoCloud. For marketplace images it will be Azure.
 	CloudInitDataSource common.CloudInitDataSource `json:"cloudInitDataSource,omitempty"`
+	//Source VM to create the image from
+	SourceVM *string `json:"sourceVM,omitempty"`
```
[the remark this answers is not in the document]
damanMSFT: Done locally. Pushing the changes in a bit.

go.mod
damanMSFT: Sure

PR #254, opened 2024-08-01
Return error when cpu or memory is nil when GPU is present

dhavalpopat:


services/compute/virtualmachine/virtualmachine.go
```diff
@@ -198,6 +198,9 @@ func (c *client) getWssdVirtualMachineHardwareConfiguration(vm *compute.VirtualM
 			}
 		}
 		if vm.HardwareProfile.VirtualMachineGPUs != nil {
+			if vm.HardwareProfile.CustomSize == nil || vm.HardwareProfile.CustomSize.CpuCount == nil || vm.HardwareProfile.CustomSize.MemoryMB == nil {
+				return nil, errors.Wrapf(errors.InvalidInput, "CPU count and MemoryMB must be specified")
```
raghavendra-nataraj, line 202: Can we add the comment  on when the CPU and MEmoryMB must be specified?

PR #263, opened 2024-08-20
VirtualHardDisk Download Support

damanMSFT:


services/compute/virtualmachine/virtualmachine.go
```diff
@@ -565,7 +565,8 @@ func (c *client) getVirtualMachineStorageProfileOsDisk(d *wssdcloudcompute.Disk)
 		return &compute.OSDisk{}
 	}
 	return &compute.OSDisk{
-		Vhd: &compute.VirtualHardDisk{URI: &d.Diskname},
+		Name: &d.Diskname,
+		Vhd:  &compute.VirtualHardDisk{URI: &d.Diskname},
```
msftpgayam, line 569: remove VHD URI

msftpgayam (approved):

services/compute/galleryimage/client.go
huanyushao, line 1: func (c *GalleryImageClient) CreateOrUpdate(ctx context.Context, location, imagePath, name string, compute *compute.GalleryImage) (*compute.GalleryImage, error) {
    if compute != nil && compute.GalleryImageProperties != nil {
        // If SourceVirtualMachine is not empty, set SourceType to VMOSDISK_SOURCE
        if compute.GalleryImageProperties.SourceVirtualMachine != "" {
            compute.SourceType = common.ImageSource_VMOSDISK_SOURCE
        } else {
            compute.SourceType = common.ImageSource_LOCAL_SOURCE
        }
    }
    return c.internal.CreateOrUpdate(ctx, location, imagePath, name, compute)
}

Can you fix the CreateOrUpdate function as well so that moc-operator can call CreateOrUpdate directly with fixed imagePath.

services/storage/virtualharddisk/virtualharddisk.go
```diff
@@ -82,6 +87,8 @@ func getVirtualHardDisk(c *wssdcloudstorage.VirtualHardDisk, group string) *stor
 			HyperVGeneration:    c.HyperVGeneration,
 			DiskFileFormat:      c.DiskFileFormat,
 			ContainerName:       &c.ContainerName,
+			SourcePath:          &c.SourcePath,
+			SourceType:          c.SourceType,
```
huanyushao, line 91: I think we should remove this. Otherwise the cloudagent is exposing the vhd's download url.

damanMSFT: Good Point. Removed.

services/compute/virtualmachine/virtualmachine.go
damanMSFT: I think we need to confirm that this won't affect any teams. This could be fixed in another PR and do a test run on it to make sure removing it is not breaking anything.

msftpgayam: Sounds good

msftpgayam (dismissed):

msftpgayam:
Could you add a comment about the use case of DiskFileFormat_Unknown and how that doesn't affect the existing scenarios.

msftpgayam (approved):

services/compute/virtualmachine/virtualmachine.go
```diff
@@ -607,6 +607,7 @@ func (c *client) getVirtualMachineStorageProfileOsDisk(d *wssdcloudcompute.Disk)
 		}
 	}
 	return &compute.OSDisk{
+		Name:        &d.Diskname,
```
raghavendra-nataraj, line 610: Why cant we just use VHD? Why do we need the name? Since the vallues are the same

raghavendra-nataraj (dismissed):

services/compute/virtualmachine/virtualmachine.go
damanMSFT: The idea here is to remove it from the VHD.URI because it doesn't makes sense to have the name there. I wanted to be sure that it is not affecting any other team before doing it. So will do it as part of a seperate PR later and run the tests.

services/storage/storage.go
```diff
@@ -34,12 +34,17 @@ type VirtualHardDiskProperties struct {
...
 	HyperVGeneration common.HyperVGeneration `json:"hyperVGeneration,omitempty"`
-	//DiskFileFormat - File format of the disk. possible values are common.DiskFileFormat_DiskFileFormatVHD and common.DiskFileFormat_DiskFileFormatVHDX
+	//DiskFileFormat - File format of the disk. possible values are common.DiskFileFormat_DiskFileFormatVHD, common.DiskFileFormat_DiskFileFormatVHDX and common.DiskFileFormat_DiskFileFormatUNKNOWN.
+	//If not specified, default value is common.DiskFileFormat_DiskFileFormatVHDX
 	DiskFileFormat common.DiskFileFormat `json:"diskFileFormat,omitempty"`
 	// CloudInitDataSource - The cloud init data source to be used with the image. [NoCloud, Azure]. Default Value – NoCloud. For marketplace images it will be Azure.
 	CloudInitDataSource common.CloudInitDataSource `json:"cloudInitDataSource,omitempty"`
 	// Container name
 	ContainerName *string `json:"containername,omitempty"`
+	// Source path for the VHD. Only supports url.
```
madhanrm, line 44: We dont have to expose these in SDK. Can you just pass this in the Function call and pass it to the proto?

madhanrm (dismissed):

services/storage/virtualharddisk/wssd.go
```diff
@@ -116,7 +117,7 @@ func getVirtualHardDiskPrecheckRequest(group, container string, vhds []*storage.
 	return request, nil
 }
 
-func getVirtualHardDiskRequest(opType wssdcloudcommon.Operation, group, container, name string, storage *storage.VirtualHardDisk) (*wssdcloudstorage.VirtualHardDiskRequest, error) {
+func getVirtualHardDiskRequest(opType wssdcloudcommon.Operation, group, container, name string, storage *storage.VirtualHardDisk, sourcePath string, sourceType common.ImageSource) (*wssdcloudstorage.VirtualHardDiskRequest, error) {
```
liwonMS, line 120: I wonder if you need to update the function signature for precheck too for this type of create.

liwonMS: Have you talked to @JamesLiang29 he was implementing precheck foir VHD. https://github.com/microsoft/wssdcloudagent/pull/1029/files

services/storage/storage.go
```diff
@@ -34,7 +34,8 @@ type VirtualHardDiskProperties struct {
 	Statuses map[string]*string `json:"statuses"`
 	//HyperVGeneration - Gets the HyperVGenerationType of the VirtualMachine created from the image. Possible values are common.HyperVGeneration_HyperVGenerationV1 and common.HyperVGeneration_HyperVGenerationV2
 	HyperVGeneration common.HyperVGeneration `json:"hyperVGeneration,omitempty"`
-	//DiskFileFormat - File format of the disk. possible values are common.DiskFileFormat_DiskFileFormatVHD and common.DiskFileFormat_DiskFileFormatVHDX
+	//DiskFileFormat - File format of the disk. possible values are common.DiskFileFormat_DiskFileFormatVHD, common.DiskFileFormat_DiskFileFormatVHDX and common.DiskFileFormat_DiskFileFormatUNKNOWN.
+	//If not specified, default value is common.DiskFileFormat_DiskFileFormatVHDX
```
liwonMS, line 38: i dont really understand when will the UNKONWN play in role if the default is VHDX. Also I dont think there's anywhere in your code that you check for the value and substitute to default (VHDX).

damanMSFT: Discussed on a seperate chat. UNKNOWN Format is needed when a user passes in a download url and doesn't specify the file format for it. In that case we set it to unknown until we determine the format after downloading the file.

liwonMS (dismissed):

services/storage/virtualharddisk/wssd.go
damanMSFT: Reached out to him. As his cloudagent PR is not checked in yet and feature is not live will work with him to enable precheck for this scenario too.

msftpgayam (approved):

madhanrm (approved):

PR #267, opened 2024-08-31
Fix multiple issues in current PR by crating V2 APIs

danmamsft:
The current APIs have two major issues:
1. the string returned back to caller doesn't include a length, so the caller cannot use memory safe str functions which requires length. This is caught by OneFuzz: https://dev.azure.com/microsoft/OS/_workitems/edit/53371227. The fix in V2 API is to return length as well
2. the C string returned to caller (both data blob and error string) is not freed, causing memory leak. The fix is to let caller allocates the memory and pass in allocated buffer. If the buffer is not pre-allocated or is too small, the V2 API will return buffer too small with the expected size for caller to reallocate the proper size.

one improvement made to encrypt and decrypt API is made the algorithm as a parameter

Test:
I have modifies testMoc.exe in IgvmAgent (please see my draft code in branch user/dama/testwrapperfix in Igvmagent repo for reference).
Here are the test I have done:
SecurityLoginV2: tested happy path; wrong/empty login file path; invalid/empty group name
CrateKeyV2: tested happy path to create both AES and RSA key; invalid/empty group name
KeyvaultKeyDecryptDataV2: tested happy path; buffer too small case (caller call API twice to get decrypt working); invalid algorithm; invalid groupName/keyvaultName; empty input data
KeyvaultKeyEncryptDataV2: same test coverage as decrypt.Also tested Encrypt and decrypt together and made sure decrypted data matches original plaintext
KeyvaultKeyCreateOrUpdateV2: tested happy path to create AES key and RSA key; tested invalid algorithm;invalid/empty groupName/keyvaultName
KeyvaultGetPublicKeyV2: tested happy path; tested return key empty; invalid/empty groupName/keyvaultName/keyName

Tested the onefuzz repro, with the fix, original error went away.

Tested with replacing igvmagent in a haas cluster. TVM is stopping and starting with no issue.

### DIFF
--- a/wrapper/cpp/main.go
+++ b/wrapper/cpp/main.go
@@ -11,11 +11,20 @@
 
 package main
 
+/*
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+typedef unsigned long long uint64_t;
+
+*/
 import "C"
 
 import (
 	"context"
 	"time"
+	"unsafe"
 
 	"github.com/microsoft/moc-sdk-for-go/services/security/authentication"
 	"github.com/microsoft/moc-sdk-for-go/services/security/keyvault"
@@ -24,6 +33,28 @@ import (
 	"github.com/microsoft/moc/pkg/auth"
 	"github.com/microsoft/moc/pkg/config"
 )
+
+// possible Win32 return values
+const (
+	Win32Succeed int = 0               // ERROR_SUCCESS
+	Win32ErrorInsufficientBuffer int = 122    // ERROR_INSUFFICIENT_BUFFER
+	Win32ErrorBadArg int = 160         // ERROR_BAD_ARGUMENTS
+	Win32ErrorFunctionFail int = 1627  // ERROR_FUNCTION_FAILED
+)
+
+// helper function to best effort copy over the error message
+func copyErrorMessage(errMessage *C.char, errMessageBuffer *C.char, errMessageSize C.ulonglong) {
+	if (errMessage != nil) {
+		if (errMessageBuffer != nil && errMessageSize > 0) {
+			msglength := C.strlen(errMessage)
+			if (errMessageSize < msglength) {
+				msglength = errMessageSize
+			}
+			C.strncpy(errMessageBuffer, errMessage, msglength)
+		}
+		C.free(unsafe.Pointer(errMessage))
+	}
+}
 
 // This function exists to maintain backwards compatability. Please use SecurityLoginCV.
 //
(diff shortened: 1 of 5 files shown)


wrapper/cpp/main.go
```diff
@@ -11,11 +11,20 @@
 
 package main
 
+/*
```
SethBe, line 14: Nit can you add a comment pointing out that this is not actually a comment lol?

wrapper/cpp/main.go
```diff
@@ -63,6 +94,46 @@ func SecurityLoginCV(serverName *C.char, groupName *C.char, loginFilePath *C.cha
 	return nil
 }
 
+//export SecurityLoginV2
```
SethBe, line 97: nit might be good to add the calling pattern, as a comment.

wrapper/cpp/main.go
```diff
@@ -63,6 +94,46 @@ func SecurityLoginCV(serverName *C.char, groupName *C.char, loginFilePath *C.cha
...
+		copyErrorMessage(C.CString(telemetry.FilterSensitiveData(err.Error())), errMessageBuffer, errMessageSize)
+		return C.int(Win32ErrorFunctionFail)
+	}
+
+	//Provide moc version information after login
+	telemetry.EmitWrapperTelemetry("SecurityLoginV2", C.GoString(cv), "", "", C.GoString(serverName))
+	return C.int(Win32Succeed)
+}
+
 // This function exists to maintain backwards compatability. Please use KeyvaultKeyEncryptDataCV.
```
SethBe, line 137: >CV. [](http://example.com/codeflow?start=94&length=3)

update all these to v2

wrapper/cpp/main.go
```diff
@@ -101,6 +172,60 @@ func KeyvaultKeyEncryptDataCV(serverName *C.char, groupName *C.char, keyvaultNam
...
+	}
+
+	response, err := keyClient.Encrypt(ctx, C.GoString(groupName), C.GoString(keyvaultName), C.GoString(keyName), parameters)
+	if err != nil {
+		telemetry.EmitWrapperTelemetry("KeyvaultKeyEncryptDataV2", C.GoString(cv), err.Error(), "keyClient.Encrypt", C.GoString(serverName))
+		copyErrorMessage(C.CString(telemetry.FilterSensitiveData(err.Error())), errMessageBuffer, errMessageSize)
+		return C.int(Win32ErrorFunctionFail)
+	}
+
+	// retrun base64 encoded string
```
SethBe, line 211: > retrun [](http://example.com/codeflow?start=3&length=7)

return :)

wrapper/cpp/main.go
```diff
@@ -101,6 +172,60 @@ func KeyvaultKeyEncryptDataCV(serverName *C.char, groupName *C.char, keyvaultNam
...
+//export KeyvaultKeyEncryptDataV2
+func KeyvaultKeyEncryptDataV2(serverName *C.char, groupName *C.char, keyvaultName *C.char, keyName *C.char, input *C.char, algorithm *C.char, cv *C.char, timeoutInSeconds C.int, outputBuffer *C.char, outputBufferSize *C.ulonglong, errMessageBuffer *C.char, errMessageSize C.ulonglong) C.int {
+	if (serverName == nil || groupName == nil || keyvaultName == nil || keyName == nil || input == nil || algorithm == nil || cv == nil || outputBufferSize == nil) {
+		copyErrorMessage(C.CString("Invalid Argument"), errMessageBuffer, errMessageSize)
+		telemetry.EmitWrapperTelemetry("KeyvaultKeyEncryptDataV2", C.GoString(cv), "", "InvalidArgument", C.GoString(serverName))
+		return C.int(Win32ErrorBadArg)
+	}
+
+	keyClient, err := getKeyvaultKeyClient(C.GoString(serverName), C.GoString(cv))
+	// if errror occurs, return an empty string so that caller can tell between error and encrypted blob
```
SethBe, line 184: >// if errror occurs, return an empty string so that caller can tell between error and encrypted blob [](http://example.com/codeflow?start=1&length=100)

outdated comment?

wrapper/cpp/main.go
```diff
@@ -101,13 +187,87 @@ func KeyvaultKeyEncryptDataCV(serverName *C.char, groupName *C.char, keyvaultNam
...
+// return:
+//   - int: the win32 code. If output buffer is not big enough this function will return Win32ErrorInsufficientBuffer which is win32 error ERROR_INSUFFICIENT_BUFFER.
+//     Caller when seeing this error should allocate outputBuffer to size outputBufferSize
+// Calling pattern:
+//   call KeyvaultKeyEncryptDataV2 for the first time. If returned error Win32ErrorInsufficientBuffer, caller allocate outputBuffer to size outputBufferSize and call KeyvaultKeyEncryptDataV2 again.
+//
+//export KeyvaultKeyEncryptDataV2
+func KeyvaultKeyEncryptDataV2(serverName *C.char, groupName *C.char, keyvaultName *C.char, keyName *C.char, input *C.char, algorithm *C.char, cv *C.char, timeoutInSeconds C.int, outputBuffer *C.char, outputBufferSize *C.ulonglong, errMessageBuffer *C.char, errMessageSize C.ulonglong) C.int {
+	if (serverName == nil || groupName == nil || keyvaultName == nil || keyName == nil || input == nil || algorithm == nil || cv == nil || outputBufferSize == nil) {
+		copyErrorMessage(C.CString("Invalid Argument"), errMessageBuffer, errMessageSize)
```
SethBe, line 213: >C.CString("Invalid Argument") [](http://example.com/codeflow?start=19&length=29)

do we need to free this
?

SethBe (approved):
:shipit:

wrapper/cpp/main.go
danmamsft: it is freed in copyErrorMessage

raghavendra-nataraj (approved):

wrapper/cpp/main.go
```diff
@@ -332,6 +598,80 @@ func KeyvaultGetPublicKeyCV(serverName *C.char, groupName *C.char, keyvaultName
...
+		telemetry.EmitWrapperTelemetry("KeyvaultGetPublicKeyV2", C.GoString(cv), "", "EmptyKey", C.GoString(serverName))
+		copyErrorMessage(C.CString("Returned key is empty"), errMessageBuffer, errMessageSize)
+		return C.int(Win32ErrorFunctionFail)
+	}
+
+	pemPkcs1KeyPub := (*keys)[0].Value
+	return handleOutputBuffer(serverName, *pemPkcs1KeyPub, cv, outputBuffer, outputBufferSize, "KeyvaultGetPublicKeyV2")
+}
+
+func handleOutputBuffer(serverName *C.char, resultGoString string, cv *C.char, outputBuffer *C.char, outputBufferSize *C.ulonglong, funcName string) C.int {
```
SethBe, line 654: >handleOutputBuffer(s [](http://example.com/codeflow?start=5&length=20)

Uber nit: Consider moving this to the top with the other helpers and add comment.

SethBe (approved):
:shipit:

SethBe (approved):
:shipit:

PR #275, opened 2024-10-05
added grpc error parsing interceptor

nikohobart:
Updates moc tag and adds client-side interceptor.

Details and testing in this PR: https://github.com/microsoft/moc/pull/298


msftpgayam (dismissed):

go.mod
```diff
@@ -7,19 +7,26 @@ require (
 	github.com/Azure/go-autorest/autorest v0.11.29
 	github.com/Azure/go-autorest/autorest/date v0.3.0
 	github.com/google/uuid v1.6.0
-	github.com/microsoft/moc v0.23.1
+	github.com/microsoft/moc v0.23.2
+	github.com/microsoft/moc-pkg v0.23.2
```
apluchik, line 11: Does it make sense to add a dependency on moc-pkg if this is an open source repo?
@msftpgayam should this be in moc?

nikohobart: Oh sorry didn't realize there was a strong distinction between the two. The main reason I added the interceptor to moc-pkg was because we already have server-side interceptors in that repo.

Would also be happy to move this client-side interceptor to moc though because upgrading moc -> moc-pkg -> wssd-sdk-for-go -> wssdcloudagent is a pretty painful process and reducing the number of layers there would be convenient for future changes.

yeonghunki (approved):

PR #280, opened 2024-10-17
added priority field to VM

connievi:


bgriddaluru (commented):
For using the new field. Refer to 'getWssdVirtualMachine' . Basically, there are wrapper functions to convert fields from structs in 'moc' pkg to local defined structs (ex: compute.go) and vicecersa.

PR #286, opened 2024-10-25
Add sleep between VM stop and start for restart command

ganeshamohana:
https://msazure.visualstudio.com/One/_workitems/edit/30026680

### DIFF
--- a/services/compute/virtualmachine/client.go
+++ b/services/compute/virtualmachine/client.go
@@ -88,6 +88,9 @@ func (c *VirtualMachineClient) Restart(ctx context.Context, group string, name s
 	if err != nil {
 		return
 	}
+	//wait for 5 seconds before starting the VM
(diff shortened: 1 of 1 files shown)


services/compute/virtualmachine/client.go
```diff
@@ -88,6 +88,9 @@ func (c *VirtualMachineClient) Restart(ctx context.Context, group string, name s
 	if err != nil {
 		return
 	}
+	//wait for 5 seconds before starting the VM
```
msftpgayam, line 91: Just thinking out loud, I wonder if it makes sense to have a loop that waits for the powerstate to change to OFF before sending the Start request, instead of a blind sleep.

ganeshamohana: We can create a loop to monitor the power state that also has a timeout.

PR #293, opened 2024-11-08
Add sleep between VM stop and start for restart command

ganeshamohana:
https://msazure.visualstudio.com/One/_workitems/edit/30026680/


nikohobart (approved):
Approving but 5 seconds seems like a lot for merging into prod, I guess it's a temporary necessity for successful restart cmds?

Would one alternative be to add a check at the "stop" level in nodeagent that confirms the vm has actually stopped? Sorry I'm missing some context

PR #308, opened 2025-04-03
Adding observability around cert renew

djshah19:


services/security/authentication/casigned.go
```diff
@@ -70,57 +69,62 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
 	return &response.Token, nil
 }
 
-func renewRoutine(ctx context.Context, group, server string) {
+func renewRoutine(ctx context.Context, group, server string, logger logr.Logger) {
```
raghavendra-nataraj, line 72: Can we add a interface for implementing the logging, Since we need this for moccli too

djshah19: I addressed this

services/security/authentication/casigned.go
```diff
@@ -70,57 +69,63 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
...
-		log.Println("Certificate renewal complete")
+		logger.Info("Certificate renewal complete")
 	}
 }
 
 // Get methods invokes the client Get method
 func (c *client) LoginWithConfig(ctx context.Context, group string, loginconfig auth.LoginConfig, enableRenewRoutine bool) (*auth.WssdConfig, error) {
+	logger := log.FromContext(ctx) // Retrieve the logger from context
+	if logger.GetSink() == nil {   // Check if the logger has a valid sink
+		logger = log.Log // Use the global default logger as a fallback
```
raghavendra-nataraj, line 122: We should use the default logger created in MOC as the fallback

services/security/authentication/casigned.go
```diff
@@ -70,57 +69,63 @@ func (c *client) Login(ctx context.Context, group string, identity *security.Ide
...
+	}
 
+	logger.Info("Generating client CSR")
```
raghavendra-nataraj, line 125: These are going to be printed in moccti. We should limit the logger to the go routine. This function should return the appropriate errors rather than logging

services/security/authentication/casigned.go
djshah19: done!

services/security/authentication/casigned.go
djshah19: done!

PR #316, opened 2025-04-29
[DRAFT] Add support for force delete which does not check if VM exists upon VM Delete

han-sogawa-microsoft:


madhanrm:
Can I understand the intention behind this force. Is there a discussion that happened, as to why we are going in this direction and spending cycles here?

PR #324, opened 2025-05-29
[CMP] add support for token authentication between cloud and node agent

wecha-ms-v2:
this PR contains the change to support mapping the new nodeagentauthmode parameter between node and cloud agent


seanyen (approved):
LGTM

PR #331, opened 2025-07-22
Fix unwrap and decrypt for callers who don't specify a version

jirenugo:
Fixing a bug which would cause unwrap and decrypt for callers that don't specify a keyversion. If the data was encrypted to an older version, the SDK will still pick the latest version. The PR removes the `Get` calls to the cloudagent since the cloudagent already validates the key before operations (`wrap`, `unwrap`, `encrypt`, `decrypt`, `rotate`)

Validation:
- [x] Unit tests
- [x] wssdtesting


jirenugo:
YAML tests run 
https://msazure.visualstudio.com/msk8s/_build/results?buildId=132027694&view=results

madhanrm:
Auto-closing this PR as part of the monthly stale-PR cleanup - it has been open for over 90 days or has had no activity in the last 90 days. This is reversible: reactivate the PR if it is still needed. - automated cleanup on behalf of @madhanm

PR #367, opened 2026-01-20
Add sdnConfig to lnet/vnet

liwonMS:
### Description

This PR extends the MOC Go SDK to surface SDN (Software Defined Networking) configuration on logical network and virtual network resources, enabling downstream consumers to read Network Controller state from these objects.

#### What this PR does

- **Adds `NetworkControllerConfig` struct** (`services/network/network.go`) — A new type that captures the SDN configuration state with five boolean fields: `IsSdnEnabled`, `IsSdnVnetEnabled`, `IsSdnVnetV2Enabled`, `IsSdnLBV2Enabled`, and `IsLegacySdnEnabled`.

- **Adds `NetworkControllerConfig` field** to both `LogicalNetwork` and `VirtualNetwork` structs, making SDN configuration accessible on both resource types.

- **Adds protobuf-to-SDK conversion** — Implements `getNetworkControllerConfig()` helper functions in both the logical network and virtual network service packages to convert the protobuf `NetworkControllerConfig` message into the SDK type.

- **Updates `moc` dependency** to the version containing the new `NetworkControllerConfig` protobuf definitions.

#### Why

Operators and controllers need to determine the SDN configuration of a network (e.g., whether SDN is enabled, which SDN version is in use) to make informed decisions about network provisioning and IPAM behavior. This change makes that information available through the standard Go SDK.

Ref:

image


services/network/network.go
```diff
@@ -599,6 +613,8 @@ type VirtualNetwork struct {
 	Tags map[string]*string `json:"tags"`
 	// VirtualNetworkProperties - Properties of the virtual network.
 	*VirtualNetworkPropertiesFormat `json:"properties,omitempty"`
+	// NetworkControllerConfig - The network controller configuration
+	NetworkControllerConfig *NetworkControllerConfig `json:"networkControllerConfig,omitempty"`
```
madhanrm, line 617: MOC is not windows specific stack. Exposing NetworkController configuration is NOT OKAY.

madhanrm: Second, the API is expected to be consistent with Azure ARM API. That is our design principle. Please do not break that by adding arbitrary properties like this.

madhanrm:
Auto-closing this PR as part of the monthly stale-PR cleanup - it has been open for over 90 days or has had no activity in the last 90 days. This is reversible: reactivate the PR if it is still needed. - automated cleanup on behalf of @madhanm

PR #382, opened 2026-05-11
Add registeredIPAddresses (reserve specific IPs in IPAM without NIC allocation)

yingzhan-msft:
Add registeredIPAddresses (reserve specific IPs in IPAM without NIC allocation)

Add SDK support for the new registeredIPAddresses RPC on both
LogicalNetwork and VirtualNetwork clients.

Changes:
- services/network/registeredips/: new shared package holding the SDK-side
  types for the feature so both LNet and VNet packages can consume the same
  surface without duplicating definitions.
  - SubnetRegisteredIPs { SubnetName, RegisteredIPAddresses[] }: snapshot
    of registered IPs on a subnet (used as both input and return type).
  - IPAddressUpdateFailure { SubnetName, IPAddress, Code, Error }.
  - IPUpdateErrorCode named-type wrapper around the proto enum so we can
    hang String/MarshalJSON/MarshalYAML methods without leaking the
    generated rpc/common import into SDK consumers.
  - SubnetsToProto / SubnetsFromProto / FailuresFromProto converters.

- services/network/{logicalnetwork,virtualnetwork}/registered_ips.go:
  per-network-type wrappers re-exporting the shared types and implementing
  UpdateRegisteredIPs against the LNet / VNet gRPC client.

- services/network/{logicalnetwork,virtualnetwork}/client.go:
  Service interface + LogicalNetworkClient / VirtualNetworkClient now expose
  UpdateRegisteredIPs with named returns
  (subnetPersistedIPs, failures, err) so the type-symmetric signature
  documents intent in godoc.

- services/network/{logicalnetwork,virtualnetwork}/{logical,virtual}network.go:
  RegisteredIPAddresses round-trips on the API model so list/show output
  reflects the persisted state.


services/network/logicalnetwork/client.go
```diff
@@ -16,6 +16,7 @@ type Service interface {
 	CreateOrUpdate(context.Context, string, string, *network.LogicalNetwork) (*network.LogicalNetwork, error)
 	Delete(context.Context, string, string) error
 	Precheck(ctx context.Context, location string, logicalNetworks []*network.LogicalNetwork) (bool, error)
+	UpdateRegisteredIPs(ctx context.Context, locationName, name string, subnetRegisteredIPs []SubnetRegisteredIPs) (subnetPersistedIPs []SubnetRegisteredIPs, failures []IPAddressUpdateFailure, err error)
```
madhanrm, line 19: Why are we making this change?

There is no such use case to expose such functionality.
At this point we are not making any feature changes in MOC

liwonMS (dismissed):

jeanlag1 (approved):

liwonMS (approved):

PR #393, opened 2026-08-06
Add unit test for DiskAttach idempotency

raghavendra-nataraj:
## What
Adds a Go unit test for `VirtualMachineClient.DiskAttach` idempotency, driving it through a fake `Service` (no live cloud agent):

- **disk already attached to the VM** → returns `nil` (no-op success) and issues **no** `CreateOrUpdate`
- **disk not attached** → appended and persisted via `CreateOrUpdate`
- **attaching a different disk** → preserves the already-attached one

Regression guard for the fix in #392 (return `nil` instead of `AlreadyExists` when the disk is already attached to the target VM). Without the fix, the first subtest fails (it would get `AlreadyExists`).

## Stacked on #392
This PR is **based on `natarajr/diskattach-idempotent`** (the fix) so the test compiles against the fixed behavior and CI is green. GitHub will auto-retarget it to `main` once #392 merges. Please merge #392 first.

## Note on CI coverage
The repo's `make all` → `unittest` target only runs `./pkg/client/...` and `./services/security/...`, so this package isn't executed by the default CI gate. The test runs via `make test` / `go test ./services/compute/virtualmachine/`. Verified passing locally (`-vet=off` to skip an unrelated pre-existing vet warning at `virtualmachine_test.go:48`).


raghavendra-nataraj:
Closing — the unit test has been folded into the fix PR #392 (commit 974328b). The separate test PR is the wssdtesting Pester BVT instead.

